PR #8, opened 2018-07-20
Write a MEE specification

palmer-dabbelt:
I just quickly threw something together here, but I figure it's better than nothing.


palmer-dabbelt:
Closed in favor of https://github.com/sifive/freedom-mee/pull/13

PR #9, opened 2018-07-21
Declare inline functions as extern

palmer-dabbelt:
Without this we get undefined references to these symbols when GCC
decides not to inline them.  I'd avoid declaring them as "static inline"
because I didn't want to increase code size too much.


palmer-dabbelt:
This doesn't work.

PR #10, opened 2018-07-23
Allow HiFive1 PLL to change operating frequency

nategraff-sifive:
Includes a new test program named pll_set_hz which demonstrates the new functionality the PLL driver


Makefile.am
```diff
@@ -112,6 +111,13 @@ hello_CFLAGS         += -ffunction-sections -fdata-sections
 hello_CFLAGS         += -march=$(FRAMEWORK_BOARD_DTS_MARCH) -mabi=$(FRAMEWORK_BOARD_DTS_MABI)
 hello_LDFLAGS         = -L. -Wl,--gc-sections -Wl,-Map=hello.map
 
+check_PROGRAMS		 += pll_set_hz
+pll_set_hz_SOURCES	  = test/pll_set_hz.c
+pll_set_hz_CFLAGS     = -mmee-machine=@MACHINE_NAME@
+pll_set_hz_CFLAGS    += -ffunction-sections -fdata-sections
+pll_set_hz_CFLAGS    += -march=$(FRAMEWORK_BOARD_DTS_MARCH) -mabi=$(FRAMEWORK_BOARD_DTS_MABI)
+pll_set_hz_LDFLAGS    = -L. -Wl,--gc-sections -Wl,-Map=pll_set_hz.map
+
```
palmer-dabbelt: There's some broken alignment in this block.

dts/hifive1.dts
```diff
@@ -179,7 +180,7 @@
                         compatible = "sifive,spi0";
                         interrupt-parent = <&plic>;
                         interrupts = <6>;
-                        reg = <0x10014000 0x1000 0x20000000 0x20000000>;
+                        reg = <0x10014000 0x1000 0x20400000 0xF42400>;
```
palmer-dabbelt: The base address was actually more correct before, but the size was wrong.

mee/clock.h
```diff
@@ -14,10 +14,17 @@ struct __mee_clock_vtable {
     long (*set_rate_hz)(struct mee_clock *clk, long hz);
 };
 
+typedef void (*mee_clock_pre_rate_change_callback)(void *priv);
 typedef void (*mee_clock_rate_change_callback)(void *priv);
```
palmer-dabbelt: Can you rename these "post_rate_change", to keep it uniform with the "pre_rate_change"?

mee/clock.h
```diff
@@ -28,15 +35,27 @@ inline long mee_clock_get_rate_hz(const struct mee_clock *clk) { return clk->vta
 /* Attempts to set the current rate of the given clock to as close as possible
  * to the given rate.  Returns the actual value that's been selected, which
  * could be anything! */
-inline long mee_clock_set_rate_hz(struct mee_clock *clk, long hz)
+static inline long mee_clock_set_rate_hz(struct mee_clock *clk, long hz)
```
palmer-dabbelt: I just opened #11 that should fix this in a better way.

mee/clock.h
```diff
@@ -28,15 +35,27 @@ inline long mee_clock_get_rate_hz(const struct mee_clock *clk) { return clk->vta
...
+
     if (clk->rate_change_callback != NULL)
         clk->rate_change_callback(clk->rate_change_callback_priv);
+
     return out;
 }
 
-/* Registers a callback that must be called whenever. */
+/* Registers a callback that must be called before a rate change */
+inline void mee_clock_register_pre_rate_change_callback(struct mee_clock *clk, mee_clock_pre_rate_change_callback cb, void *priv)
```
palmer-dabbelt: Also needs updating for #11.

src/drivers/sifive,fe310-g000,pll.c
```diff
@@ -15,6 +18,156 @@
...
+        MEE_ACCESS_ONCE(pllcfg) &= ~(PLL_SEL);
+
+    /* Make sure we're running off of the external oscillator for stability */
+    if(pll->pllref != NULL)
+        MEE_ACCESS_ONCE(pllcfg) |= PLL_REFSEL;
+
+    /* Configure the PLL to run at the requested init frequency.
+     * Using the vtable instead of the user API because we want to control
+     * when the callbacks occur. */
+    pll->vtable->clock.set_rate_hz(&(pll->clock), pll->init_rate);
```
palmer-dabbelt: You should allow `pll->init_rate` to be 0, in case a user just wants to use whatever is baked into the hardware.

src/drivers/sifive,fe310-g000,pll.c
```diff
@@ -49,10 +202,115 @@ long __mee_driver_sifive_fe310_g000_pll_get_rate_hz(const struct mee_clock *cloc
...
         return pllout;
 
-    return pllout / (2 << MEE_GET_FIELD(div, DIV_DIV));
+    return pllout / (2 * (MEE_GET_FIELD(div, DIV_DIV) + 1));
+}
+
+static inline long abs(long val)
+{
+    return (val >= 0 ? val : val * -1);
+}
```
palmer-dabbelt: This should live somewhere as a utility function, maybe in compiler.h?

src/entry.S
```diff
@@ -54,8 +54,7 @@ _enter:
      * so we don't have to write another ABI. */
     csrr a0, mhartid
     li a1, 0
-    .weak __mee_before_main
-    la a2, __mee_before_main
+    li a2, 0
```
palmer-dabbelt: We should remove this from the bootloader spec as well, it wasn't a good idea.

test/pll_set_hz.c
```diff
@@ -0,0 +1,28 @@
...
+
+#include <stdio.h>
+#include <mee/clock.h>
+
+#define PLL_MIN_OUT 16000000
+#define PLL_MAX_OUT 320000000
+
+void main() {
+    /* TODO: have DTS parser generate handle for PLL? */
+    struct mee_clock * pll = &(__mee_dt_clock_4.clock);
```
palmer-dabbelt: We need a better interface here, as this will only work on the HiFive1 (and even then, only with the device tree exactly as it is currently specified).  Instead we should add functionality into the MEE for enumerating cores and then finding the clocks that drive them.

It's fine to leave this in as a test case for now, but there should be some way of ensuring it compiles for other targets.

PR #14, opened 2018-11-01
Mee timer driver support

bsousi5:


mee/drivers/sifive,timer.h
```diff
@@ -0,0 +1,30 @@
...
+#include <mee/compiler.h>
+#include <mee/timer.h>
+
+struct __mee_driver_vtable_timer {
+    struct mee_timer_vtable vtable;
+};
+
+void __mee_driver_sifive_timer_init(struct mee_timer *timer, int timebase);
+int  __mee_driver_sifive_timer_get(struct mee_timer *timer, unsigned long *tv);
+int  __mee_driver_sifive_timebase_get(struct mee_timer *timer, int *tb);
```
palmer-dabbelt: Both the timebase and the ticks should be a "uint64_t", as that's what the RISC-V ISA mandates.

Makefile.am
```diff
@@ -97,10 +100,13 @@ libriscv__mmachine__@MACHINE_NAME@_a_SOURCES = \
 	src/drivers/sifive,fe310-g000,pll.c \
 	src/drivers/sifive,test0.c \
 	src/drivers/sifive,gpio0.c \
+	src/drivers/sifive,timer.c \
```
palmer-dabbelt: Currently there is no "sifive,timer" device in the devicetree, and I don't think it makes a lot of sense to add one because it's not an independent hardware block.  IIRC the timers are actually provided by the CLINT, so it should be possible to register the CLINT as a timer driver.

mee/times.h
```diff
@@ -0,0 +1,16 @@
+/* Copyright 2018 SiFive, Inc */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#ifndef MEE__TIMES_H
+#define MEE__TIMES_H
```
palmer-dabbelt: I think these should just be in `mee/timer.h`, as they're really not their own interface.

src/drivers/sifive,timer.c
```diff
@@ -0,0 +1,45 @@
...
+{
+    struct __mee_driver_timer *mee_timer = (void *)(timer);
+    unsigned long val = 0;
+
+    #if __riscv_xlen == 32
+        unsigned int hi, hi1, lo;
+
+        asm volatile ("csrr %0, mcycleh" : "=r"(hi));
+        asm volatile ("csrr %0, mcycle" : "=r"(lo));
+        asm volatile ("csrr %0, mcycleh" : "=r"(hi1));
```
palmer-dabbelt: Ah, if the timer is from `mcycle` then this driver should really be a driver for a hart.  Since the timebase could differ between harts, then the interface should contain a way to obtain a timer structure that corresponds to each hart.

src/drivers/sifive,timer.c
```diff
@@ -0,0 +1,45 @@
...
+
+int __mee_driver_sifive_timebase_get(struct mee_timer *timer, int *tb)
+{
+    struct __mee_driver_timer *mee_timer = (void *)(timer);
+
+    if (mee_timer->timebase) {
+        *tb = mee_timer->timebase;
+	return 0;
+    } else {
+        return -1;
```
palmer-dabbelt: If `-1` indicates an error then the NOP functions should return `-1` not `0`.

mee/times.h
```diff
@@ -0,0 +1,16 @@
...
+
+/* Read the machine cycle counts from csrr register */
+int mee_get_cyclecount(unsigned long *mcc);
```
palmer-dabbelt: Since these functions are public they need to be documented as part of the API docs in `doc/`.

src/times.c
```diff
@@ -0,0 +1,43 @@
...
+    return mee_timer_get(__MEE_DT_TIMER_HANDLE, mcc);
+}
+
+int mee_get_timebase_frequency(int *timebase)
+{
+    return mee_timebase_get(__MEE_DT_TIMER_HANDLE, timebase);
+}
+
+#ifndef __MEE_DT_TIMEBASE_FREQUENCY
+#define __MEE_DT_TIMEBASE_FREQUENCY 1000000
```
palmer-dabbelt: I don't think it's good to have a default here.

PR #18, opened 2018-12-13
 Add --with-bultin-libgloss, to build a libgloss

palmer-dabbelt:
We sometimes ship toolchains with an MEE libgloss, but not always.
This allows users to build libgloss directly into libriscv__menv__mee.a,
which may be a cleaner solution.  Note that this libgloss includes crt0
as part of the mee.a, which may not be the correct thing to do.


palmer-dabbelt:
This is blocked by https://github.com/sifive/freedom-mee/pull/16

PR #21, opened 2018-12-17
Add driver for sifive,spi0

nategraff-sifive:


mee/spi.h
```diff
@@ -0,0 +1,47 @@
...
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#ifndef MEE__SPI_H
+#define MEE__SPI_H
+
+struct mee_spi;
+
+struct mee_spi_config {
+    enum {
+        SPI_SINGLE,
```
palmer-dabbelt: These should be within the MEE namespace.

doc/mee-spec.md
```diff
@@ -124,6 +124,44 @@ Sets the baud rate of the given UART.  Returns 0 on success, or -1 on
...
+
+The SPI interface allows users to perform transfers over the SPI protocol.
+
+`int mee_spi_init(struct mee_spi *spi, int baud_rate)`
+
+Initializes the given SPI. This must be called exactly once before any other
+function in this SPI is called. It is invalid to initialize a SPI more than
+once.
+
+`int mee_spi_transfer(struct mee_spi *spi, struct mee_spi_config *config, int words, char *tx_buf, char *rx_buf)`
```
palmer-dabbelt: `int words` should be a `size_t`.  Also, calling it `words` is a bit confusing as it's a number of bytes, right?

nategraff-sifive: I initially called it "words" because I was initially going to support a variable-length frame size. I'll rename it to `len`.

mee/spi.h
nategraff-sifive: You mean prepend `MEE_` to these, right?

nategraff-sifive:
After talking to @bsousi5 I added `mee_spi_get_device()`

palmer-dabbelt:
@nategraff-sifive It looks like there's some conflicts, can you rebase?

src/drivers/sifive,spi0.c
```diff
@@ -0,0 +1,232 @@
+/* Copyright 2018 SiFive, Inc */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include <metal/drivers/sifive,spi0.h>
+#include <metal/io.h>
+
+/* Register byte offsets */
+#define SPI_REG_SCKDIV  0x000
```
bsousi5: Even though they are local, can we also prefix these with METAL_, like other drivers.

PR #23, opened 2018-12-19
Add support for ITIM

nategraff-sifive:
Adds support to `gloss/crt0.S` for copying the `.itim` section into the ITIM memory.

Adds `mee/itim.h` for library support for interacting with the ITIM. To start with, define the macro `MEE_PLACE_IN_ITIM` which decorates functions to place in the ITIM.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -79,6 +79,7 @@ nobase_include_HEADERS = \
 	mee/drivers/sifive,test0.h \
 	mee/compiler.h \
 	mee/clock.h \
+	mee/itim.h \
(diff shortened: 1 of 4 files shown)


install-sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 # install - install a program, script, or datafile
 
-scriptversion=2016-01-11.22; # UTC
+scriptversion=2014-09-12.12; # UTC
```
palmer-dabbelt: This whole file looks like noise.

PR #30, opened 2019-01-11
Add Support for PMP Configuration (No vtable version)

nategraff-sifive:
Driver for PMP configuration which uses __riscv_xlen to determine which pmpcfgX CSRs to configure without requiring vtable lookups.


nategraff-sifive:
I should document the public interface in Doxygen before merge as well.

PR #34, opened 2019-01-21
Document the MEE public API with Doxygen

nategraff-sifive:
Generate HTML documentation for the MEE public API with Doxygen

Docs are still being fleshed out. Feel free to push commits into this PR.


nategraff-sifive:
Close in favor of #42

PR #46, opened 2019-01-31
CLIC interrupts enabled and serviced

bsousi5:


mee/interrupt.h
```diff
@@ -86,9 +99,22 @@ inline int mee_interrupt_disable(struct mee_interrupt *controller, int id)
     return controller->vtable->interrupt_disable(controller, id);
 }
 
-/* Internal-use utility function for a given interrupt controller */
+/* Enable an interrupt vector for a given controller and id */
```
nategraff-sifive: Add Doxygen comments for new public API functions.

PR #53, opened 2019-02-14
Define a public API for GPIO output

nategraff-sifive:
I feel like the right public interface is on a per-pin basis, so I wrapped the existing internal driver to provide that.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,6 +91,7 @@ nobase_include_HEADERS = \
 	metal/clock.h \
 	metal/compiler.h \
 	metal/cpu.h \
+	metal/gpio.h \
 	metal/interrupt.h \
 	metal/io.h \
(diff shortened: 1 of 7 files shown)


nategraff-sifive:
@palmer-dabbelt Can you clarify what you mean by "link"?

palmer-dabbelt:
`__metal_driver_led_on()` (https://github.com/sifive/freedom-metal/blob/master/src/drivers/sifive%2Cgpio-leds.c#L28) should call `__metal_driver_sifive_gpio0_output_set()` (https://github.com/sifive/freedom-metal/blob/master/src/drivers/sifive%2Cgpio0.c#L30).

nategraff-sifive:
Buttons and Switches will need to wait for a GPIO input API, but I updated LEDs to use this API.

PR #56, opened 2019-02-21
Create a public atomic lock API

### LINKED ISSUE #76: Add atomic locks
Depends on #75 so that we can detect when locks have been allocated in a region of memory which doesn't support atomic operations.

nategraff-sifive:
@palmer-dabbelt  This will fail to compile on targets without the atomic extension with an error like:

```
freedom-e-sdk/freedom-metal/src/lock.c: Assembler messages:
freedom-e-sdk/freedom-metal/src/lock.c:18: Error: unrecognized opcode `amoswap.w.aq a5,a5,(a4)'
freedom-e-sdk/freedom-metal/src/lock.c:29: Error: unrecognized opcode `amoswap.w.rl x0,x0,(a5)'
```

What do you think is the correct solution here? If there's a way to detect the arch with the preprocessor that would probably be ideal.

Closes #76

### DIFF
--- a/metal/lock.h
+++ b/metal/lock.h
@@ -35,7 +35,7 @@ struct metal_lock {
  * If the lock cannot be initialized, attempts to take or give the lock
  * will result in a Store/AMO access fault.
  */
-int metal_lock_init(struct metal_lock *lock);
+inline int metal_lock_init(struct metal_lock *lock);
 
 /*!
  * @brief Take a lock
@@ -45,7 +45,7 @@ int metal_lock_init(struct metal_lock *lock);
  * If the lock initialization failed, attempts to take a lock will result in
  * a Store/AMO access fault.
  */
-int metal_lock_take(struct metal_lock *lock);
+inline int metal_lock_take(struct metal_lock *lock);
 
 /*!
  * @brief Give back a held lock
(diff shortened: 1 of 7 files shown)


palmer-dabbelt:
`__riscv_atomic` does it.  The set of pre-defined preprocessor macros lives here: https://gcc.gnu.org/svn/gcc/trunk/gcc/config/riscv/riscv-c.c

palmer-dabbelt:
Also, @palmer isn't me :)

nategraff-sifive:
@palmer-dabbelt We need the memory enumeration API anyway. We should use it to check at lock initialization if the lock is properly allocated in a memory region that supports atomic operations.

metal/lock.h
```diff
@@ -0,0 +1,60 @@
...
+
+/*!
+ * @brief Take a lock
+ * @param lock The handle for a lock
+ * @return 0 if the lock is successfully taken
+ *
+ * If the lock initialization failed, attempts to take a lock will result in
+ * a Store/AMO access fault.
+ */
+int metal_lock_take(struct metal_lock *lock);
```
palmer-dabbelt: You should make the implementations available in the header for inlining, the locks are small enough that it's usually smaller code size to inline them than to call the function.

nategraff-sifive:
Depends on #77

PR #57, opened 2019-02-23
Add support for HTIF interfaces

palmer-dabbelt:
This just supports HTIF console (always via the write syscall) and
shutdown (without an exit code).  With this I can run code on Spike.


bsousi5:
I only saw fromhost getting initialize to 0, or reset. But what set this value?
Unless it is getting update by Spike given that it is a global variable.

PR #65, opened 2019-03-08
Make crt0.S compatible with RV32E

nategraff-sifive:
Replaces the t3 register with a0 so that crt0.S is compatible with both
the I and E base integer ISA.


dbarbi1:
boy this is some sharp looking code to me ;)

PR #69, opened 2019-03-23
Enable UART RX and fix getc

mmicko:
This makes receiving data to actually work. Tested on HiFive 1 board.


mmicko:
Complete example using serial send/receive is micropython port at https://github.com/mmicko/micropython/tree/experimental/ports/freedom-e

palmer-dabbelt:
Thanks a bunch for doing this!

PR #70, opened 2019-03-27
Enable reads from stdin through metal_tty_getc()

nategraff-sifive:
Someone should double-check that I implemented `_read()` properly.

### DIFF
--- a/gloss/sys_read.c
+++ b/gloss/sys_read.c
@@ -1,9 +1,25 @@
+#include <metal/tty.h>
 #include <sys/types.h>
 #include <errno.h>
+#include <unistd.h>
 
 ssize_t
 _read(int file, void *ptr, size_t len)
 {
-  errno = ENOSYS;
-  return -1;
+  if (file != STDIN_FILENO) {
+    errno = ENOSYS;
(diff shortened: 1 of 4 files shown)


gloss/sys_read.c
```diff
@@ -1,9 +1,25 @@
...
-  errno = ENOSYS;
-  return -1;
+  if (file != STDIN_FILENO) {
+    errno = ENOSYS;
+    return -1;
+  }
+
+  char c = metal_tty_getc();
+
+  /* If we got back -1, no default input device is available */
```
bsousi5: Since length is one of this function argument, shouldnt we be reading until of line/CR?

nategraff-sifive: If I keep reading until CR/LF then this doesn't work for `getchar()`, it just loops reading until it receives a CR, which is not the behavior of `getchar()` on Unix, where it reads a single character without waiting for a CR.

nategraff-sifive:
I'm testing this more and I'm not happy with the behavior of the system either way I implement `_read()`. Definitely save this for after v201903.

PR #72, opened 2019-03-28
Make pll_configs const

mmicko:
This change saves us 0x880 (2176) bytes of RAM.


nategraff-sifive:
Great catch, thanks!

Could you mark the `pll_config_t` argument to `get_pll_config_freq()` and `configure_pll()` as `const` as well to fix the  the "discards `const` qualifier from pointer target type" warnings?

mmicko:
Sure, should be fine now.

PR #77, opened 2019-04-03
Create memory enumeration API

nategraff-sifive:


metal/memory.h
```diff
@@ -0,0 +1,81 @@
...
+struct _metal_memory_attributes {
+	int R : 1;
+	int W : 1;
+	int X : 1;
+	int C : 1;
+	int A : 1;
+};
+
+/*!
+ * @brief A handle for a memory
```
bsousi5: Is it worthwhile to note that it is a memory 'block' in the description.
Memory is rather vague.

PR #81, opened 2019-04-16
Add API for dropping to a lower privilege mode

nategraff-sifive:
I added a struct for the entire register file, but for right now the privilege drop only sets `sp` and `ra`.

### DIFF
--- a/src/privilege.c
+++ b/src/privilege.c
@@ -1,0 +1,57 @@
+/* Copyright 2019 SiFive, Inc */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include <stddef.h>
+
+#include <metal/privilege.h>
+
+#define METAL_MSTATUS_MIE_OFFSET	3
+#define METAL_MSTATUS_MPIE_OFFSET	7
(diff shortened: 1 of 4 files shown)


src/privilege.c
```diff
@@ -0,0 +1,57 @@
...
+#define METAL_MSTATUS_MPIE_OFFSET	7
+#define METAL_MSTATUS_SIE_OFFSET	1
+#define METAL_MSTATUS_SPIE_OFFSET	5
+#define METAL_MSTATUS_UIE_OFFSET	0
+#define METAL_MSTATUS_UPIE_OFFSET	4
+
+#define METAL_MSTATUS_MPP_OFFSET	11
+#define METAL_MSTATUS_MPP_MASK		3
+
+void metal_privilege_drop_to_mode(enum metal_privilege_mode mode,
```
bsousi5: A nit thing, why "xxx_drop_to_mode" and not "xxx_change_mode". The later is more obvious IMHO, as mode can be change up and down

nategraff-sifive: You can't go up in privilege, you can only drop privilege. You can only return to a high privilege mode through an exception or interrupt.

PR #82, opened 2019-04-18
Fix inlining of Metal Lock API

nategraff-sifive:
Declaring the functions inline in the header and defining them in the
source file was breaking the linker. Define and declare the functions in
the header and declare them extern inline in the source file.

### DIFF
--- a/metal/lock.h
+++ b/metal/lock.h
@@ -4,10 +4,16 @@
 #ifndef METAL__LOCK_H
 #define METAL__LOCK_H
 
+#include <metal/memory.h>
+#include <metal/compiler.h>
+
 /*!
  * @file lock.h
  * @brief An API for creating and using a software lock/mutex
  */
+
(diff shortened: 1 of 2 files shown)


metal/lock.h
```diff
@@ -4,11 +4,16 @@
...
 
+#include <metal/memory.h>
+#include <metal/compiler.h>
+
 /*!
  * @file lock.h
  * @brief An API for creating and using a software lock/mutex
  */
 
+#define _METAL_STORE_AMO_ACCESS_FAULT 7
```
bsousi5: This is already define in metal_exception_code_e;
May be metal_exception_code_e; should be brought out to interrupt.h

nategraff-sifive: `metal_exception_code_e` is defined in `metal/drivers/riscv,cpu.h`. Should I include that file?

PR #85, opened 2019-05-13
Update timer

reclusejack:
1. update _times(), return ticks based on different bsps.
2. update _gettimeofday(), add tv_usec support.


palmer-dabbelt:
@bsousi5 

````

/* Structure describing CPU time used by a process and its children.  */
struct tms
  {
    clock_t tms_utime;          /* User CPU time.  */
    clock_t tms_stime;          /* System CPU time.  */

    clock_t tms_cutime;         /* User CPU time of dead children.  */
    clock_t tms_cstime;         /* System CPU time of dead children.  */
  };
````

PR #87, opened 2019-05-14
Update metal to use inline functions produce in metal-header

bsousi5:
Metal Headers generated by the Freedom-devicetree-tools will no longer produce fields in its structures to capture the const value described by the design dts file. Instead it will generate inline functions to retrieve to those values. This also allow us to avoid using ".weak" attributes for its structure declaring.

This PR capture the changes for this migration.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -73,6 +73,7 @@ nobase_include_HEADERS = \
 	metal/drivers/sifive,spi0.h \
 	metal/drivers/sifive,test0.h \
 	metal/drivers/sifive,uart0.h \
+	metal/machine/inline.h \
 	metal/machine/platform.h \
 	metal/button.h \
 	metal/clock.h \
@@ -85,7 +86,6 @@ nobase_include_HEADERS = \
 	metal/led.h \
 	metal/lock.h \
 	metal/machine.h \
-	metal/machine-inline.h \
 	metal/memory.h \
 	metal/pmp.h \
 	metal/privilege.h \
@@ -101,7 +101,7 @@ metal/machine.h: @MACHINE_HEADER@
 	@mkdir -p $(dir $@)
 	cp $< $@
 
-metal/machine-inline.h: @MACHINE_INLINE@
+metal/machine/inline.h: @MACHINE_INLINE@
 	@mkdir -p $(dir $@)
 	cp $< $@
 
@@ -142,7 +142,7 @@ libriscv__mmachine__@MACHINE_NAME@_a_CCASFLAGS = @MENV_METAL@ @MMACHINE_MACHINE_
 # This will generate these sources before the compilation step
 BUILT_SOURCES = \
 	metal/machine.h \
-	metal/machine-inline.h \
+	metal/machine/inline.h \
 	metal/machine/platform.h \
 	riscv__menv__metal.specs \
 	riscv__mmachine__@MACHINE_NAME@.specs
--- a/Makefile.am
+++ b/Makefile.am
@@ -85,6 +85,7 @@ nobase_include_HEADERS = \
 	metal/led.h \
 	metal/lock.h \
 	metal/machine.h \
+	metal/machine-inline.h \
 	metal/memory.h \
 	metal/pmp.h \
 	metal/privilege.h \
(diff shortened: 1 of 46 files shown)


src/drivers/fixed-clock.c
```diff
@@ -7,16 +7,22 @@
 
 #include <metal/drivers/fixed-clock.h>
 #include <stddef.h>
+#include <metal/machine.h>
 
 long __metal_driver_fixed_clock_get_rate_hz(const struct metal_clock *gclk)
 {
     const struct __metal_driver_fixed_clock *clk = (void *)gclk;
-    return clk->rate;
+    return __metal_driver_fixed_clock_rate();
```
palmer-dabbelt: This is missing the clock pointer.

src/drivers/fixed-factor-clock.c
```diff
@@ -7,22 +7,26 @@
...
+    struct metal_clock *parent = __metal_driver_fixed_factor_clock_parent(gclk);
     long parent_rate = 1;
-    if(clk->parent) {
-        parent_rate = clk->parent->vtable->get_rate_hz(clk->parent);
+    if(parent) {
+        parent_rate = parent->clock->get_rate_hz(parent);
     }
 
-    return clk->mult * parent_rate / clk->div;
+    return __metal_driver_fixed_factor_lock_mult() * parent_rate / __metal_driver_fixed_factor_clock_div();
```
palmer-dabbelt: This is `lock` instead of `clock`.

src/drivers/riscv,clint0.c
```diff
@@ -8,31 +8,34 @@
 #include <metal/io.h>
 #include <metal/cpu.h>
 #include <metal/drivers/riscv,clint0.h>
+#include <metal/machine.h>
 
 unsigned long long __metal_clint0_mtime_get (struct __metal_driver_riscv_clint0 *clint)
 {
     __metal_io_u32 lo, hi;
+    unsigned long control_base = __metal_driver_sifive_clint0_control_base((struct metal_interrupt *)clint);
```
palmer-dabbelt: This shouldn't require casting.

metal/compiler.h
```diff
@@ -4,9 +4,11 @@
...
 #define METAL__COMPILER_H
 
-#define __METAL_DECLARE_VTABLE(type)                        \
-    asm(".weak " #type);                                  \
-    const struct type type                                \
+#define __METAL_DECLARE_VTABLE(type)                      \
+    extern const struct type type;			  
+
+#define __METAL_DEFINE_VTABLE(type)                       \
+    const struct type type
```
palmer-dabbelt: I thought the goal here was to get rid of vtables entirely?

src/drivers/fixed-factor-clock.c
bsousi5: thanks will correct it

src/drivers/riscv,clint0.c
bsousi5: thanks will remove

metal/compiler.h
bsousi5: Without vtable, we would need to expose each node structure handle/address all the way to the top metal APIs. I was with the impression that we don't want to do that.

src/drivers/fixed-clock.c
bsousi5: You right, should add one since custom and/or future design way have many clocks to select from

Makefile.am
```diff
@@ -86,6 +85,7 @@ nobase_include_HEADERS = \
 	metal/led.h \
 	metal/lock.h \
 	metal/machine.h \
+	metal/machine-inline.h \
```
nategraff-sifive: Should we make it `metal/machine/inline.h` so that it matches `metal/machine/platform.h`? Or maybe change `platform.h` to match this one?

metal/drivers/sifive,spi0.h
```diff
@@ -14,30 +14,13 @@ struct __metal_driver_vtable_sifive_spi0 {
     const struct metal_spi_vtable spi;
 };
 
-struct __metal_driver_sifive_spi0;
+//struct __metal_driver_sifive_spi0;
```
nategraff-sifive: Should this be deleted instead of commented?

metal/drivers/sifive,test0.h
```diff
@@ -11,19 +11,12 @@ struct __metal_driver_vtable_sifive_test0 {
     const struct __metal_shutdown_vtable shutdown;
 };
 
-struct __metal_driver_sifive_test0;
+//struct __metal_driver_sifive_test0;
```
nategraff-sifive: Another

PR #89, opened 2019-05-15
Remove indirection from metal/machine.h

nategraff-sifive:
Copy the machine header directly into `metal/machine.h` instead of `metal/machine/@MACHINE_NAME@.h`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -54,8 +54,6 @@ endif # PRECONFIGURED
 # Install some METAL-specific headers, one of which is automatically generated.
 # The files that aren't automatically generated are the same for all machines.
 nobase_include_HEADERS = \
-	metal/machine/@MACHINE_NAME@.h \
-	metal/machine/platform.h \
 	metal/drivers/fixed-clock.h \
 	metal/drivers/fixed-factor-clock.h \
 	metal/drivers/riscv,clint0.h \
@@ -75,6 +73,7 @@ nobase_include_HEADERS = \
 	metal/drivers/sifive,spi0.h \
 	metal/drivers/sifive,test0.h \
 	metal/drivers/sifive,uart0.h \
+	metal/machine/platform.h \
 	metal/button.h \
 	metal/clock.h \
 	metal/compiler.h \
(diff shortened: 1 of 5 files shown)


sifivekevin:
I checked out the 'master' branch of freedom-e-sdk, then updated the freedom-metal submodule to this remove-indirection branch.  When I try to create a standalone-project in Freedom Studio I'm getting this error:

```
checking for riscv64-unknown-elf-ar... riscv64-unknown-elf-ar
checking the archiver (riscv64-unknown-elf-ar) interface... ar
checking dependency style of riscv64-unknown-elf-gcc... gcc3
configure: error: in `/c/workspaces/fs-runtime/e31-arty-ri/bsp/build/debug':
configure: error: --with-platform-header is required with the --with-preconfigured option
See `config.log' for more details
make: *** [Makefile:218: /c/workspaces/fs-runtime/e31-arty-ri/bsp/build/debug/Makefile] Error 1
```

sifivekevin:
Switching freedom-metal back to "v201903" restores my ability to create a standalone project.

nategraff-sifive:
@sifivekevin This will only work for the development-19.05 branch, since there are a number of changes throughout E SDK and freedom-devicetree-tools that need to be updated concurrently.

nategraff-sifive:
Does this also fix the problem you're seeing with running `clean`, or does that needs its own fix?

sifivekevin:
That needs a different fix, but that fix lay in Freedom Studio code I believe.  We currently add the ephemeral 'install' folder to the list of folders (it is the only item in the list currently) to examine for headers.    So the index fails after a clean (and also fails when the project is first created, before it get built).   I think we need to add the more permanent folders in the "freedom-metal" tree instead.

nategraff-sifive:
Adding the folders in the freedom-metal tree will fix some of those errors, but `metal/machine.h` and `metal/machine/platform.h` are generated by the freedom-metal build and will only exist in `install/include`, so those two will still have problems.

sifivekevin:
Thinking in the context of a standalone project,  what could a user do to cause either of those files to change in subsequent builds?  In other words, how stable are these headers in a standalone project?

sifivekevin:
Im thinking, is it feasible to make additional copies of these files in folder that does not get removed during a clean, for the sole purpose of maintaining an indexable project.  The freedom-metal makefile can simply replace them as needed.

PR #91, opened 2019-05-21
Use __metal_boot_hart in crt0.S

nategraff-sifive:
The hartid matching `__metal_boot_hart`
- Executes the initialization code and constructors in `_start`
- Is chosen by the default `secondary_main()` to run `main()`

Also, fix the mtimecmp_set behavior for CLINT and CLIC so that they set the time compare of the requested hart instead of assuming hartid = 0.


gloss/crt0.S
```diff
@@ -11,6 +11,8 @@
 
 /* crt0.S: Entry point for RISC-V METAL programs. */
 
+#include <metal/machine/platform.h>
```
bsousi5: I am not sure it is a good to include this.
Up to now, crt0.S is only depend the linker file I believe

PR #94, opened 2019-05-24
Adding asm support for new tests

dconn-sifive:


bsousi5:
I don't think this is the right way to add this, as it will cause confusion and heavy maintenance headache in the long run. As is, developers can and will include this file directly into the metal drivers.

Since it is completely independent and some what separate from metal, may be the best way forward is to have its own example-asm-tests repo, which it does not need to include metal.h, and such.
It will be submodules into freedom-e-sdk, and can be build through freedom-e-sdk.

dconn-sifive:
So let's do a self contained test example which can be added as a submodule - that's fine.  However there are two changes to the entry code that we still need (fcsr=0 and reading mstatus in early_trap_vector) so I'll do those on a separate PR.

PR #96, opened 2019-05-30
fcsr=0 needed for default value, read mstatus in early_trap_vector

dconn-sifive:


src/entry.S
```diff
@@ -81,6 +81,7 @@ _enter:
 .align 2
 early_trap_vector:
     .cfi_startproc
+    csrr t0, mcause
```
nategraff-sifive: Maybe it'd also be valuable to read `mepc` and `mtval`?

PR #97, opened 2019-05-31
Add documentation for how to configure the FE310-G000 PLL

nategraff-sifive:


dconn-sifive:
Can we reference __libc_init_array in crt0.s?

bsousi5:
Yes, it is being called in crt0.s.

This is a reserved libc function.

nategraff-sifive:
`__libc_init_array` is a private symbol, it shouldn't be part of the public API documentation.

bsousi5:
Oh I missed understood the question.

David,
What do you mean by referencing it?

Bunnaroath

dconn-sifive:
B - what I was getting at was how is the constructor used, how is it invoked, etc?  Customers will have this question.  Also I think it will be useful to add a note on how to update the frequency in the metal.h, not just reference the design.dts node.  

Something like "to update the startup frequency, change the .init_rate field in the hfclk node"

/* From clock@4 */
struct __metal_driver_sifive_fe310_g000_pll __metal_dt_clock_4 = {
    .vtable = &__metal_driver_vtable_sifive_fe310_g000_pll,
    .clock.vtable = &__metal_driver_vtable_sifive_fe310_g000_pll.clock,
/* From clock@3 */
    .pllsel0 = &__metal_dt_clock_3.clock,
/* From clock@1 */
    .pllref = &__metal_dt_clock_1.clock,
/* From prci@10008000 */
    .divider_base = &__metal_dt_prci_10008000,
    .divider_offset = 12UL,
/* From prci@10008000 */
    .config_base = &__metal_dt_prci_10008000,
    .config_offset = 8UL,
    .init_rate = 16000000UL,
};

nategraff-sifive:
The constructor is invoked by libgloss, the only people who should want to change how that happens are those replacing libgloss.

Changing metal.h is definitely not a supported interface. It's absolutely possible, and probably works in the field, but everything in there is internal interface and is subject to change at any time.

dconn-sifive:
What would be the recommended way to have a customer get to their target frequency (outside of the default 16 MHz)?

nategraff-sifive:
The recommended way is to edit the DeviceTree and re-run freedom-devicetree-tools.

bsousi5:
As Nate indicated, the constructor is so embedded with crt0.s/startup codes, they would need to replace the gloss.
We may have to look into how to provide a formal APIs to shim custom startup codes to call metal APIs.
But for now, the metal lipgloss would server as an example.

As for PLL setting API, we can see if we can added API for user to set/override with their own setting, without changing the DTS/Metal.h file.

Bunnaroarth

PR #98, opened 2019-06-04
Rename CPU internal method from timer_get to mcycle_get

nategraff-sifive:
Only changes the internal method name in the CPU vtable

### DIFF
--- a/metal/cpu.h
+++ b/metal/cpu.h
@@ -20,7 +20,7 @@ struct metal_cpu;
 typedef void (*metal_exception_handler_t) (struct metal_cpu *cpu, int ecode);
 
 struct metal_cpu_vtable {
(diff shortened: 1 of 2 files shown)


metal/cpu.h
```diff
@@ -69,7 +69,7 @@ int metal_cpu_get_num_harts();
  * @return The value of the CPU cycle count timer
  */
 inline unsigned long long metal_cpu_get_timer(struct metal_cpu *cpu)
-{ return cpu->vtable->timer_get(cpu); }
```
bsousi5: The public metal_cpu_get_timer() API should also be rename to metal_cpu_cycle_count()

nategraff-sifive: That breaks public API, should we include the old function and mark it deprecated?

PR #99, opened 2019-06-05
Qspi debug

Jaxonwht:
Modified `src/drivers/sifive,spi0.c` only. Changed some of the preprocessor directives to match those specified in [SiFive FE310-G002 Manual v19p05](https://sifive.cdn.prismic.io/sifive%2F59a1f74e-d918-41c5-b837-3fe01ba7eaa1_fe310-g002-manual-v19p05.pdf).
 
Another thing to note is that I moved the CS deassert line to only after the RXDADA FIFO has been dequeued. There are some timing issues (according to Logic analyzer) if the CS# was immediately deasserted after the last byte in TXDATA was enqueued in the sending FIFO. Placing the deassert line so late in the process probably hurt performance a bit. I need to be able to get the SCK frequency in order to calculate a better latency, and this relates to the next point.

Any `set_baud_rate` doesn't seem to modify SPI clock rates at all. The analyzer showed the clock rate stayed approximately at 4MHz whether I set `baud_rate` to 10 or 100,000.

`METAL_SPI_REGW(METAL_SPI_REG_FCTRL) = METAL_SPI_CONTROL_IO` switches the SPI from memory-mapped to programmable I/O mode. I have to either comment this line to start gdb normally and then manually edit the register value through gdb before I start the program, or I have to comment out the `flash bank my_first_flash fespi 0x40000000 0 0 0 $_TARGETNAME 0x20004000` line in `openocd.cfg`. I'm not familiar with openocd to give a better fix.

### DIFF
--- a/src/drivers/sifive_spi0.c
+++ b/src/drivers/sifive_spi0.c
@@ -7,6 +7,7 @@
 #include <metal/drivers/sifive_spi0.h>
 #include <metal/io.h>
 #include <metal/machine.h>
+#include <time.h>
 
 /* Register fields */
 #define METAL_SPI_SCKDIV_MASK         0xFFF
@@ -52,16 +53,16 @@ static int configure_spi(struct __metal_driver_sifive_spi0 *spi, struct metal_sp
 {
     long control_base = __metal_driver_sifive_spi0_control_base((struct metal_spi *)spi);
     /* Set protocol */
-    METAL_SPI_REGW(METAL_SIFIVE_SPI0_FMT) &= ~(METAL_SPI_PROTO_MASK);
+    METAL_SPI_REGB(METAL_SIFIVE_SPI0_FMT) &= ~(METAL_SPI_PROTO_MASK);
     switch(config->protocol) {
         case METAL_SPI_SINGLE:
-            METAL_SPI_REGW(METAL_SPI_REG_FMT) |= METAL_SPI_PROTO_SINGLE;
+            METAL_SPI_REGB(METAL_SIFIVE_SPI0_FMT) |= METAL_SPI_PROTO_SINGLE;
             break;
         case METAL_SPI_DUAL:
-            METAL_SPI_REGW(METAL_SPI_REG_FMT) |= METAL_SPI_PROTO_DUAL;
+            METAL_SPI_REGB(METAL_SIFIVE_SPI0_FMT) |= METAL_SPI_PROTO_DUAL;
             break;
         case METAL_SPI_QUAD:
-            METAL_SPI_REGW(METAL_SPI_REG_FMT) |= METAL_SPI_PROTO_QUAD;
+            METAL_SPI_REGB(METAL_SIFIVE_SPI0_FMT) |= METAL_SPI_PROTO_QUAD;
             break;
         default:
             /* Unsupported value */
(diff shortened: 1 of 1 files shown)


src/drivers/sifive,spi0.c
```diff
@@ -122,43 +140,81 @@ int __metal_driver_sifive_spi0_transfer(struct metal_spi *gspi,
...
+    time_t endwait; 
+
+    for (int i = 0; i < len; i++) {
+        /* Wait for RXFIFO to not be empty, but break the nested loops if timeout, this timeout method 
+         * needs refining, preferably taking into account the device specs */
+        endwait = time(NULL) + METAL_SPI_RXDATA_TIMEOUT;
+        while((rxdata = METAL_SPI_REGW(METAL_SPI_REG_RXDATA)) & METAL_SPI_RXDATA_EMPTY) {
+            if (time(NULL) > endwait) {
+                /* if timeout, immediately stop receiving more bytes */
+                goto end_receive;
```
nategraff-sifive: If the receive times out, the method should report an error with a nonzero return code.

@bsousi5, is this how we want to do timeouts in Metal? I just want to make sure this is the best way.

src/drivers/sifive,spi0.c
```diff
@@ -122,43 +140,81 @@ int __metal_driver_sifive_spi0_transfer(struct metal_spi *gspi,
...
-
-    for(int i = 0; i < len; i++) {
-        /* Wait for TXFIFO to not be full */
-        while(METAL_SPI_REGW(METAL_SPI_REG_TXDATA) & METAL_SPI_TXDATA_FULL) ;
+    METAL_SPI_REGW(METAL_SPI_REG_CSMODE) &= ~(METAL_SPI_CSMODE_MASK);
+    METAL_SPI_REGW(METAL_SPI_REG_CSMODE) |= METAL_SPI_CSMODE_HOLD;
 
-        /* Transfer byte */
-        if(tx_buf) {
-            METAL_SPI_REGW(METAL_SPI_REG_TXDATA) = tx_buf[i];
```
nategraff-sifive: Did simply writing the byte not work? It shouldn't be necessary to make writing `txdata` more complicated.

Maybe you can replace `METAL_SPI_REGW()` with`METAL_SPI_REGB()`?

src/drivers/sifive,spi0.c
```diff
@@ -122,43 +140,81 @@ int __metal_driver_sifive_spi0_transfer(struct metal_spi *gspi,
...
-        } else {
-            METAL_SPI_REGW(METAL_SPI_REG_TXDATA) = 0;
-        }
+    /* Increase the maximum delay between two consecutive frames, in units of clock cycles 
+     * The purpose is to prevent the spi controller from deasserting the cs pin in between
+     * consecutive frames. By default 0 frame interval is allowed between frames. The delay
+     * between each iteration in the for loop is more than one clock cycle. 0xA means the
+     * maximum allowed inter-frame interval is 10 clock cycles.*/
+    /** METAL_SPI_REGW(METAL_SPI_REG_DELAY1) &= ~(0xFF << METAL_SPI_INTERVAL_SHIFT); */
+    /** METAL_SPI_REGW(METAL_SPI_REG_DELAY1) |= (0xF << METAL_SPI_INTERVAL_SHIFT); */
```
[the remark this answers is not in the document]
Jaxonwht: Now I think the delay control is no longer needed because I moved the deasserting time.

src/drivers/sifive,spi0.c
```diff
@@ -122,43 +140,81 @@ int __metal_driver_sifive_spi0_transfer(struct metal_spi *gspi,
                                       char *tx_buf,
                                       char *rx_buf)
 {
+    /* Transmit and receive buffer must not be empty */
+    if (tx_buf == NULL || rx_buf == NULL) {
+        return 1;
+    }
```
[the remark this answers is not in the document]
Jaxonwht: I agree with the `rx_buf == NULL` part. However, about the `tx_buf  == null` situation, is the user just going to send `len` bytes of 0's? That is treated as dummy bytes by most spi devices.

src/drivers/sifive,spi0.c
```diff
@@ -113,6 +122,15 @@ static int configure_spi(struct __metal_driver_sifive_spi0 *spi, struct metal_sp
     /* Set CS line */
     METAL_SPI_REGW(METAL_SPI_REG_CSID) = config->csid;
 
+    /* Toggle off memory-mapped SPI flash mode, toggle on programmable IO mode
+     * It seems that with this line uncommented, the debugger cannot have access 
+     * to the chip at all because it assumes the chip is in memory-mapped mode. 
+     * I have to compile the code with this line commented and launch gdb,
+     * reset cores, reset $pc, set *((int *) 0x20004060) = 0, (set the flash
+     * interface control register to programmable I/O mode) and then continue
+     * Alternative, comment out the "flash" line in openocd.cfg */
```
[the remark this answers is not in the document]
Jaxonwht: Yeah, the only thing I'm unsure about is whether this issue happens to other kinds of spi devices. As long as the debugger does not assume the spi device is in memory-mapped mode, this line can remain uncommented.

src/drivers/sifive,spi0.c
Jaxonwht: `METAL_SPI_REGB` works

src/drivers/sifive,spi0.c
Jaxonwht: I'll change the return code to 1 instead.

src/drivers/sifive_spi0.c
```diff
@@ -4,92 +4,100 @@
...
 #define METAL_SPI_REGW(offset)  (__METAL_ACCESS_ONCE((__metal_io_u32 *)METAL_SPI_REG(offset)))
 
+#define METAL_SPI_RXDATA_TIMEOUT      1
+
 static int configure_spi(struct __metal_driver_sifive_spi0 *spi, struct metal_spi_config *config)
 {
     long control_base = __metal_driver_sifive_spi0_control_base((struct metal_spi *)spi);
     /* Set protocol */
-    METAL_SPI_REGW(METAL_SIFIVE_SPI0_FMT) &= ~(METAL_SPI_PROTO_MASK);
+    METAL_SPI_REGB(METAL_SIFIVE_SPI0_FMT) &= ~(METAL_SPI_PROTO_MASK);
```
nategraff-sifive: Wait. `METAL_SPI_REGB()` only changes a single byte. I was suggesting it exclusively in the context of `txdata` which only needs to update the least-significant byte, but you really should use `REGW` when you're trying to update an entire word.

src/drivers/sifive,spi0.c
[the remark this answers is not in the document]
duaneellissd: Actually you require the ability to "stuff 0" or "stuff 1" - this might be something you place in the "config" structure parameter to this function, add an element  "int stuffing_value" to the config struct.

src/drivers/sifive,spi0.c
duaneellissd: In the case of a SPI master (which this is) the master supplies the clock.

Think about how "bit-bang-gpio" SPI works, there is no such thing as a timeout.

The hardware SPI module just makes this faster, if there is a timeout - then you have a hardware (chip) problem.

In contrast, for SPI SLAVE - you might consider a very slow clock, or lack of a clock as a timeout but - the protocol does not define a timeout situation.

PR #106, opened 2019-06-20
Improve uart

Jaxonwht:
sys_read implements a circular buffer that blocks actual storing of data to memory. Each time the user types a string in the GNU screen, that string will be buffered until the user hits enter. Then that string will be appended with a '\n'. After that, the result string enters the stdin stream.

This behavior is actually closer to scanf, fgets, getc etc. on a standard pc. I did not implement non-blocking reads as of now. It seems that without ncurses, making non-block read calls on stdin does not have a clear use case. But if indeed that feature is desired, much more needs to be added. For example, the api will need to support fnctl, select, poll etc.

### DIFF
--- a/gloss/sys_read.c
+++ b/gloss/sys_read.c
@@ -3,7 +3,8 @@
 #include <errno.h>
 #include <unistd.h>
 
-/* Known Bugs:
+/*
+ * Known Bugs:
  * ======
  * The content of tty_buffer has to be filled before calling
  * metal_tty_putc(). This is because metal_tty_putc() always
@@ -22,7 +23,8 @@
  *
  * This happens after 
  * __metal_driver_sifive_uart0_txready(uart)
- * is called in src/drivers/sifive_uart0.c. */
+ * is called in src/drivers/sifive_uart0.c. 
+ */
 
 /* Parameters for the circular input buffer */
 #define TTY_BUFFER_LEN (127)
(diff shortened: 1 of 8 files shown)


src/drivers/riscv_clint0.c
```diff
@@ -176,6 +176,18 @@ int __metal_driver_riscv_clint0_command_request (struct metal_interrupt *control
 	    hartid = *(int *)data;
             __METAL_ACCESS_ONCE((__metal_io_u32 *)(control_base +
 					       (hartid * 4))) = METAL_ENABLE;
+	    /* Callers of this function assume it's blocking, in the sense that
```
bsousi5: You also need the same thing clic driver.

gloss/sys_read.c
```diff
@@ -1,9 +1,133 @@
...
+static uint8_t end_index;
+#define INCREASE_INDEX(n) (n = (n + 1) % WRAP_LEN)
+#define DECREASE_INDEX(n) (n = (n + TTY_BUFFER_LEN) % WRAP_LEN)
+#define BUFFER_LEN (end_index > start_index ? \
+                    end_index - start_index : \
+                    end_index + WRAP_LEN - start_index)
+/* This macro backs up start_index and end_index
+ * on the stack. This is only a temporary solution to the 
+ * bugs described above */
+#define BACKUP_INDICES(statement) \
```
bsousi5: SAVE_INDICES() may be more obvious

PR #112, opened 2019-07-04
drivers/riscv_plic0: Disable all interrupts on init

palmer-dabbelt:
The PLIC driver was only disabling interrupts for up to ndev devices,
but there may be more devices on the target because we shrink ndev to
avoid allocating too much memory for interrupt handlers.  This patch
forces the PLIC driver to always disable 1024 interrupts.

We should probably introduce a better scheme for looking up interrupt
handlers, but for now this will at least turn off all the interrupts on
any PLIC that has ever been configured.

I haven't tested this.

### DIFF
--- a/src/drivers/riscv_plic0.c
+++ b/src/drivers/riscv_plic0.c
@@ -3,6 +3,8 @@
 
 #include <metal/machine/platform.h>
 
+#define PLIC0_MAX_INTERRUPTS 1024
+
 #ifdef METAL_RISCV_PLIC0
 
 #include <metal/io.h>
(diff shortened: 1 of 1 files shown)


src/drivers/riscv_plic0.c
```diff
@@ -89,18 +91,25 @@ void __metal_driver_riscv_plic0_init (struct metal_interrupt *controller)
 
 	for(int parent = 0; parent < __METAL_PLIC_NUM_PARENTS; parent++) {
 	    num_interrupts = __metal_driver_sifive_plic0_num_interrupts(controller);
+            assert(num_interrupts < PLIC0_MAX_INTERRUPTS);
```
nategraff-sifive: We don't use assertions anywhere else (as far as I know), so I hesitate to start adding them unless we have a plan for how to deal with them when they fail.

src/drivers/riscv_plic0.c
```diff
@@ -89,18 +91,25 @@ void __metal_driver_riscv_plic0_init (struct metal_interrupt *controller)
...
 
-	    for (int i = 0; i < num_interrupts; i++) {
+            /* Disable every interrupt on the PLIC. */
+	    for (int i = 0; i < PLIC0_MAX_INTERRUPTS; i++) {
 		__metal_plic0_enable(plic, i, METAL_DISABLE);
 		__metal_plic0_set_priority(plic, i, 0);
-		plic->metal_exint_table[i] = NULL;
-		plic->metal_exdata_table[i].sub_int = NULL;
-		plic->metal_exdata_table[i].exint_data = NULL;
+
```
nategraff-sifive: What happens when these methods are called when `i` is greater than the number of configured interrupts in the PLIC?

PR #115, opened 2019-07-12
Add API to check for dcache and flush

bsousi5:


src/cache.c
```diff
@@ -2,7 +2,33 @@
...
+    case 0:
+#ifdef __METAL_CPU_0_DCACHE_HANDLE
+        return __METAL_CPU_0_DCACHE_HANDLE;
+#endif  
+        break;
+    case 1:
+#ifdef __METAL_CPU_1_DCACHE_HANDLE
+        return __METAL_CPU_1_DCACHE_HANDLE;
+#endif
+        break;
```
nategraff-sifive: Various supported targets have 8+1 cores, should we support up to CPU_8?

PR #121, opened 2019-07-19
Improve uart api

duaneellisql:
Adds polling functions for the UART interface.

### DIFF
--- a/src/tty.c
+++ b/src/tty.c
@@ -26,7 +26,7 @@ int metal_tty_getc(int *c)
    do {
         metal_uart_getc( __METAL_DT_STDOUT_UART_HANDLE, c );
         /* -1 means no key pressed, getc waits */
-    } while( c == -1 )
+    } while( -1 == *c )
         ;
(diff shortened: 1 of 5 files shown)


src/tty.c
```diff
@@ -2,21 +2,35 @@
...
+{
     return metal_uart_putc(__METAL_DT_STDOUT_UART_HANDLE, c);
 }
 
+int metal_tty_getc(int *c)
+{
+   do {
+        metal_uart_getc( __METAL_DT_STDOUT_UART_HANDLE, c );
+        /* -1 means no key pressed, getc waits */
+    } while( c == -1 )
```
nategraff-sifive: I'm getting `warning: comparison between pointer and integer` on this comparison. Maybe change to `*c == -1`?

duaneellisql: You are correct, I missed that when I created this from our internal stuff

duaneellisql: fixed pushed update

PR #126, opened 2019-07-25
Use __asm__ instead of asm keyword

nategraff-sifive:
The `asm` keyword is a GCC extension, so remove it for the more portable `__asm__`.


nategraff-sifive:
Superseded by #132

PR #128, opened 2019-07-27
Explicitly include interrupt.h in gpio.h

lesderid:
It uses `metal_interrupt` but doesn't include `interrupt.h`.


lesderid:
It would be great if the submodule in https://github.com/sifive/freedom-e-sdk could be updated after merging this.

PR #129, opened 2019-07-28
Fix: Fixes 'privilege' typo for machine mode

benmezger:


benmezger:
> Thanks for this catch!

It took me a while to figure-out why GCC was breaking with an undefined error 🤣

PR #133, opened 2019-07-30
Add RTC API and driver for sifive,rtc0

nategraff-sifive:


metal/rtc.h
```diff
@@ -0,0 +1,127 @@
...
+ * @brief List of RTC run behaviors
+ */
+enum metal_rtc_run_option {
+    METAL_RTC_STOP = 0,
+    METAL_RTC_RUN,
+};
+
+struct metal_rtc_vtable {
+    uint64_t (*get_rate)(const struct metal_rtc *const rtc);
+    uint64_t (*set_rate)(const struct metal_rtc *const rtc, const uint64_t rate);
```
bsousi5: any special reason these need to be 64bits?

metal/rtc.h
```diff
@@ -0,0 +1,127 @@
...
+ */
+struct metal_rtc {
+    const struct metal_rtc_vtable *vtable;
+};
+
+/*!
+ * @brief Get the rate of the RTC
+ * @return The rate in Hz
+ */
+inline uint64_t metal_rtc_get_rate(const struct metal_rtc *const rtc) {
```
bsousi5: For 32bits system, does not make sense to store or use 54bits value. May be uintptr_ptr type be more appropriate.

src/drivers/sifive_rtc0.c
```diff
@@ -0,0 +1,121 @@
...
+    count <<= 32;
+    count |= RTC_REGW(base, METAL_SIFIVE_RTC0_RTCCOUNTLO);
+
+    return count;
+}
+
+uint64_t __metal_driver_sifive_rtc0_set_count(const struct metal_rtc *const rtc, const uint64_t count) {
+    const uint64_t base = __metal_driver_sifive_rtc0_control_base(rtc);
+
+    RTC_REGW(base, METAL_SIFIVE_RTC0_RTCCOUNTHI) = (UINT_MAX & (count >> 32));
```
bsousi5: on low end cores, does RTC count regs are also 64bits wide?

metal/rtc.h
nategraff-sifive: The value is a clock rate, so I'm just making sure that large clock rates don't overflow. That said, this is an RTC API, so maybe it's inappropriate to support rates over 4.3 GHz.

src/drivers/sifive_rtc0.c
nategraff-sifive: Yes, it's a 64-bit register.

PR #137, opened 2019-08-01
Metal time

zongbox:
Add metal_time for internal use. In spi driver, it use the time() of libc, but we shouldn't expect user would link c library. It would cause the link error when user doesn't link libc.


zongbox:
Yes, make sense. I would move these stuff to metal/time.h. Thanks.

zongbox:
@nategraff-sifive  I had pushed the modification in this branch. The implementation of time was moved to the new files. :)

PR #140, opened 2019-08-01
Add to get number of PMP regions

zongbox:
Add these APIs because opensbi need these APIs to get the device information. The number of pmp regions and baud rate of UART could be hard coded in opensbi but it seem that provided by freedom-metal would be better.

### DIFF
--- a/metal/pmp.h
+++ b/metal/pmp.h
@@ -74,6 +74,11 @@ struct metal_pmp {
 struct metal_pmp *metal_pmp_get_device(void);
 
 /*!
+ * @brief Get the number of pmp regions for the hartid
+ */
+int metal_pmp_num_regions(int hartid);
+
+/*!
  * @brief Initialize the PMP
  * @param pmp The PMP device handle to be initialized
  *
--- a/src/pmp.c
+++ b/src/pmp.c
@@ -63,13 +63,18 @@ static uintptr_t _get_pmpaddr_granularity(uintptr_t address) {
     return (1 << (index + 3));
 }
 
-/* Get the number of pmp regions for the current hart */
-static unsigned int _pmp_regions() {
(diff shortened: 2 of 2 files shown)


src/uart.c
```diff
@@ -3,6 +3,24 @@
 
 #include <metal/uart.h>
 
+struct metal_uart *metal_uart_get_device(void)
+{
+#ifdef __METAL_DT_STDOUT_UART_HANDLE
+    return __METAL_DT_STDOUT_UART_HANDLE;
+#else
```
nategraff-sifive: The stdout handle should only matter to the TTY api, what we want here is to have a table of uart devices that we can access, just like you can see with the SPI, etc. The way to do that is to do a corresponding PR to freedom-devicetree-tools which adds the table.

src/uart.c
```diff
@@ -3,6 +3,24 @@
...
+    return NULL;
+#endif
+}
+
+int metal_uart_get_baud_by_device(void)
```
nategraff-sifive: I would say this should be `metal_uart_get_default_baud_by_device()`, if anything.

src/uart.c
zongbox: Yes. totally agree. I missed the TTY stuff. TTY APIs could meet my current need. I should re-sent the PR only for PMP stuff. Thanks for your advise.

src/uart.c
zongbox: Yes, this API name would be better and clear. If need, this API maybe should return the baud rate depends on the given uart device which you mention before.

zongbox:
I would close this PR, and separate the pmp to another PR. For UART part, maybe we could add the stuffs which we discussion above when we really need that.

zongbox:
@nategraff-sifive  I had pushed the modification in this branch. It only retain pmp stuff. :)

PR #142, opened 2019-08-01
Add new API to set/get interrupts threshold and priority for PLIC

bsousi5:


metal/interrupt.h
```diff
@@ -61,6 +66,15 @@ inline void metal_interrupt_init(struct metal_interrupt *controller)
     return controller->vtable->interrupt_init(controller);
 }
 
+/*!
+ * @brief Return interrupt controller label
+ *
+ * @param controller The handle for the interrupt controller
+ */
+inline char * metal_interrupt_label(struct metal_interrupt *controller)
```
nategraff-sifive: This should be more explicit about what it means. Maybe `metal_interrupt_get_controller_compatible_string()`?

src/drivers/riscv_cpu.c
```diff
@@ -141,7 +141,7 @@ void __metal_exception_handler (void) {
     		uintptr_t mtvt;
     		metal_interrupt_handler_t mtvt_handler;
 
-                asm volatile ("csrr %0, mtvt" : "=r"(mtvt));
+                asm volatile ("csrr %0, 0x307" : "=r"(mtvt));
```
nategraff-sifive: Is this to get around RISC-V toolchains without CLIC support?

We should probably `#define` this CSR number somewhere.

bsousi5: Does not make sense to add a metal-car.s file for just one register now. Will add one if more is needed.

metal/interrupt.h
bsousi5: This API has been replace with metal_interrupt_get_controller()

PR #143, opened 2019-08-05
Fix wrong determination of instruction length

zongbox:
Only '11' shouldn't be regarded as compressed instruction.


src/drivers/riscv_cpu.c
```diff
@@ -636,7 +636,8 @@ int __metal_driver_cpu_exception_register(struct metal_cpu *cpu, int ecode,
 int  __metal_driver_cpu_get_instruction_length(struct metal_cpu *cpu, uintptr_t epc)
 {
     /* Per ISA compressed instruction has last two bits of opcode set */
```
nategraff-sifive: Probably a good idea to update this comment to describe better what we're checking for.

zongbox: Yes, let me add more description for it.

zongbox:
Pushed the modification for adding more description about compressed instruction checking.

PR #161, opened 2019-08-20
Move mtvec initialization to startup code

zongbox:
These patches contain moving the mtvec initialization into startup code like pull request #159 , and also fixing the mask bits of MODE field of mtvec CSR. The encoding of mtvec is two bits only.


gloss/crt0.S
```diff
@@ -171,6 +192,11 @@ _skip_init:
   csrwi fcsr, 0
 1:
 
+  /* Set the real trap handler and use direct mode as default */
```
bsousi5: By doing the following, we would be coupling the startup code with metal.
Although this would be the preference, there are cases already that some user want to use their own.
It is best to have the startup as minimal as possible, and allow the rest for the user to add their pieces.

zongbox: Yes, that is why I move out the mtvec setting from interrupt initial function. In original case, when user calling the interrupt initial function, it would set the mtvec to __metal_exception_handler, and causes the user's trap handler be overwrite.

bsousi5: How did the user set the original, set the initial trap handler? which api was used?

zongbox:
Hi all
As my comment above. I move out the mtvec setting from interrupt initial function, because it would overwrite the user own trap handler which already set in mtvec. 

Assume user would set their own trap handler to mtvec after entering main function, we put the mtvec default setting in startup code could avoid overwriting user's trap handler, and could also jump to freedom-metal's trap handler when user didn't have own trap handler. The influence about this modification is that change the trap handler to __metal_exception_handler from early_trap_vector more earlier.

In my real case, opensbi would set it's trap handler into mtvec at the opensbi's initial flow. And the mtvec would be overwrite when I can call interrupt initial function.

It is OK to me that if there is more suitable place to set the default trap entry into mtvec. But it seems to me that we need to move out the mtvec setting in interrupt initial function.

gloss/crt0.S
zongbox: Before this patch, if user has own trap handler, they can set the mtvec value to their handler address in their program by using CSR instruction, after that, user call metal_interrupt_init to initialize interrupt controller, the mtvec would be set to  __metal_exception_handler by __metal_driver_riscv_cpu_controller_interrupt_init function.

nategraff-sifive:
I think the right thing to do here is to to use Metal's API to set up interrupt handling. Is there something you need us to add or change in order to make the Metal API work for your use case? I don't see a resolution where Metal doesn't own the value of mtvec.

zongbox:
It also makes sense that setting mtvec in Metal API. It looks like we won't expect the value of mtvec would be changed in normal case, and opensbi is a special use case on this situation.
It seems that adding a new API to set up trap handler entry is of no help in this special case, because it can set the value of mtvec again by using CSR instruction directly after calling metal_interrupt_init in opensbi.
I don't want to break the existing API just for one special case, so could we add a condition that set __metal_exception_handler to mtvec when the value of mtvec is only equal to early_trap_vector? If mtvec is not equal to early_trap_vector, that means user has own trap handler, and sets the entry to mtvec already, then we do nothing there. I'm not sure whether it makes sense or not. Do you have any idea?

bsousi5:
Rather than setting __metal_exception_handler to mtvec in __metal_driver_riscv_cpu_controller_interrupt_init, it is possible to have this done from clint and clic driver, after call the cpu interrupt controller init. This would allow the user to set their own mtvec handler. 

However, this pose a danger in that, how will one ensure the user handler is compatible with the rest of metal handling? How will one test can validate this? We'll be forced into situation where we have would to debug the cause of a problem originate in the user codes, but the symptom is showing up in metal codes. So having an API that allow the user to set their handler would open up this scenario. However, if the user is doing this via setting the CSR register directly, at least we know, because it is not early_trap or __metal_exception_handler.

In your use case, it find to do this since it will be part of the metal/opensbi codes.

zongbox:
If we move the mtvec setting to each drivers and without some check in it, maybe the mtvec would be set many times after all driver executes their initial function respectively?
This reminds me that the mtvec is not only handling interrupts, but also handling the exceptions, so whether we would specify the trap handler too late? and whether we shouldn't bind the mtvec setting to the interrupt related stuff?

bsousi5:
True, I forgot about exception handling from user registration, although early_trap will catch it.

zongbox:
Hi all,
If all you guys agree the idea that adding a condition filter for the setting mtvec in the original APIs, I would push the modification. Something like:
```
if (mtvec == (uintptr_t)&early_trap_vector) {
    __metal_controller_interrupt_vector(METAL_DIRECT_MODE,
                                                             (void *) (uintptr_t)&__metal_exception_handler);
}
```
And I would separate the early_trap_vector to another assembly file named "trap.S" or another name which you think more suitable. The reason as following:
Opensbi has own linker script, it won't use the metal.default.lds, etc. When we link the libmetal.a, there are some symbols in _enter function would lose their definition such as __global_pointer and __metal_chicken_bit.
I don't encounter this problem before, because no any places in libmetal.a use the functions which in enter.o, then the enter.o won't be linked. For now, we would use early_trap_vector symbol, so the enter.o object would be linked. Separate trap vector from enter.S, to ensure that we only link the stuff we needed.

bsousi5:
This sound good to me. And separating early-trap-vector to trap.S also fine. Just ensure __global_pointer and __metal_chicken_bit continue to be available for bare metal. It is fine if openSBI don’t need them.

zongbox:
Hi all,
I send a new PR from different branch. #167 
I would close this PR, Thanks all to review this.

PR #165, opened 2019-08-22
Update HTIF shutdown func.

JerryShih:
Return the exit code to spike.

### DIFF
--- a/src/drivers/sifive,htif0.c
+++ b/src/drivers/sifive,htif0.c
@@ -61,10 +61,17 @@ void __metal_driver_sifive_htif0_init(struct metal_uart *uart, int baud_rate)
 
 void __metal_driver_sifive_htif0_exit(const struct __metal_shutdown *sd, int code)
 {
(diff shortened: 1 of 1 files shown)


src/drivers/sifive,htif0.c
```diff
@@ -61,10 +61,17 @@ void __metal_driver_sifive_htif0_init(struct metal_uart *uart, int baud_rate)
 
 void __metal_driver_sifive_htif0_exit(const struct __metal_shutdown *sd, int code)
 {
-    while (1) {
-        fromhost = 0;
-        tohost = 1;
-    };
+    volatile uint64_t magic_mem[8];
+    magic_mem[0] = 93; // SYS_exit
```
JerryShih: The id of exit() function is 93:
Please check:
https://github.com/sifive/riscv-isa-sim/blob/f1104280e0b5507f8d161e9d039047d56babe30e/fesvr/syscall.cc#L72

PR #168, opened 2019-08-27
Enable Clint vector support and define new local interrupts APIs

bsousi5:
Define new Local Interrupts vector handler APIs as weak that can be override by user if needed.

### DIFF
--- a/metal/interrupt.h
+++ b/metal/interrupt.h
@@ -311,6 +311,146 @@ __inline__ int metal_interrupt_vector_disable(struct metal_interrupt *controller
     return controller->vtable->interrupt_vector_disable(controller, id);
 }
 
+/*!
+ * @brief Default interrupt vector handler, that can be overriden by user
+ * @param None
+ * @return None
+ */
+void __attribute__((weak, interrupt)) metal_interrupt_vector_handler(void);
+
+/*!
+ * @brief Metal Software interrupt vector handler, that can be overriden by user
+ * @param None
+ * @return None
+ */
+void __attribute__((weak, interrupt)) metal_software_interrupt_vector_handler(void);
+
+/*!
+ * @brief Metal Timer interrupt vector handler, that can be overriden by user
+ * @param None
+ * @return None
+ */
+void __attribute__((weak, interrupt)) metal_timer_interrupt_vector_handler(void);
+
+/*!
+ * @brief Metal External interrupt vector handler, that can be overriden by user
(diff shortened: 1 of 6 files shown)


metal/interrupt.h
```diff
@@ -311,6 +311,146 @@ __inline__ int metal_interrupt_vector_disable(struct metal_interrupt *controller
     return controller->vtable->interrupt_vector_disable(controller, id);
 }
 
+/*!
+ * @brief Default interrupt vector handler, that can be override by user
```
nategraff-sifive: "Can be overridden"

src/vector.S
```diff
@@ -0,0 +1,160 @@
+/* Copyright 2018 SiFive, Inc */
```
nategraff-sifive: 2019

src/drivers/riscv_cpu.c
```diff
@@ -92,6 +93,25 @@ void __metal_default_interrupt_handler (int id, void *priv) {
     metal_shutdown(200);
 }
 
+/* The metal_interrupt0_vector_handler() function can be redefined. */
```
nategraff-sifive: extra '0' in this comment

src/drivers/riscv_cpu.c
```diff
@@ -105,6 +125,20 @@ void __metal_default_sw_handler (int id, void *priv) {
     }
 }
 
+/* The metal_timer_interrupt_vector_handler() function can be redefined. */
+void __attribute__((weak, interrupt)) metal_timer_interrupt_vector_handler (void) {
```
nategraff-sifive: As discussed, just leaving here as a reminder that we should probably create a macro to help define all of these default overridable handlers, since they can be easily de-duplicated.

src/drivers/riscv_clint0.c
```diff
@@ -97,6 +101,36 @@ int __metal_driver_riscv_clint0_register (struct metal_interrupt *controller,
...
+{
+    int rc = -1;
+    struct metal_interrupt *intc = _get_cpu_intc();
+
+    if (intc) {
+        switch (mode) {
+        case METAL_VECTOR_MODE:
+        case METAL_DIRECT_MODE:
+            rc = intc->vtable->interrupt_set_vector_mode(intc, mode);
+            break;
```
nategraff-sifive: Is the switch necessary here? There aren't any other cases being handled.

PR #169, opened 2019-08-27
Support multi-core of PLIC driver

zongbox:
This patchset contains the supporting multi-core of PLIC driver. It should calculate the offset by hart context id to access the relevant region of each hart.

It would use the new stuff in metal header files which generated by freedom-devicetree-tools [PR freedom-devicetree-tools](https://github.com/sifive/freedom-devicetree-tools/pull/150).

This patchset doesn't support the multi-core of CLIC driver. In the CLIC specification, the multicore systems might have a CLIC per-core, so CLIC driver need to more other changed.

We also need to modify all test programs in freedom-e-sdk/software. All test program has own git repository, I would push the modification respectively.

### DIFF
--- a/metal/interrupt.h
+++ b/metal/interrupt.h
@@ -41,6 +41,22 @@ typedef enum metal_intr_priv_mode_ {
 } metal_intr_priv_mode;
 
 /*!
+ * @brief The bitmask of hart context
+ */
+typedef struct metal_affinity_ {
+    unsigned long bitmask;
+} metal_affinity;
+
+#define for_each_metal_affinity(bit, metal_affinity)                           \
+    for (bit = 0; metal_affinity.bitmask; bit++, metal_affinity.bitmask >>= 1)
+
+#define metal_affinity_set_val(metal_affinity, val)                            \
+    metal_affinity.bitmask = val;
+
+#define metal_affinity_set_bit(metal_affinity, bit, val)                       \
+    metal_affinity.bitmask |= ((val & 0x1) << bit);
+
+/*!
  * @brief Function signature for interrupt callback handlers
  */
 typedef void (*metal_interrupt_handler_t)(int, void *);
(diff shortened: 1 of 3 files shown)


zongbox:
OK,it is also good to me. I would add the APIs as following:
```
metal_interrupt_affinity_enable(struct metal_interrupt *controller, metal_affinity bitmask, int id);
metal_interrupt_affinity_disable(struct metal_interrupt *controller, metal_affinity bitmask, int id);
metal_interrupt_affinity_set_threshold(struct metal_interrupt *controller, metal_affinity bitmask, unsigned int threshold);
metal_interrupt_affinity_get_threshold(struct metal_interrupt *controller, int context_id);
```
These APIs are used to fix the problem of PLIC driver now. For CLIC driver, we need to add more relevant APIs for priority, level, etc.
If all you guys think the APIs is OK. I would send the PR for the implementation.

zongbox:
Push the modification about affinity APIs implementation.

metal/interrupt.h
```diff
@@ -80,6 +87,15 @@ struct metal_interrupt_vtable {
                            void *data);
     int (*mtimecmp_set)(struct metal_interrupt *controller, int hartid,
                         unsigned long long time);
+    int (*interrupt_affinity_enable)(struct metal_interrupt *controller,
+                                     metal_affinity bitmask, int id);
+    int (*interrupt_affinity_disable)(struct metal_interrupt *controller,
```
nategraff-sifive: If these methods return a `metal_affinity` then there can be fine-grained feedback instead of just an error code.

src/interrupt.c
```diff
@@ -76,26 +76,28 @@ extern __inline__ int metal_interrupt_enable(struct metal_interrupt *controller,
...
-                              unsigned int threshold);
+metal_interrupt_vector_enable(struct metal_interrupt *controller, int id);
 
-extern __inline__ unsigned int
-metal_interrupt_get_priority(struct metal_interrupt *controller, int id);
+extern __inline__ int
+metal_interrupt_vector_disable(struct metal_interrupt *controller, int id);
 
 extern __inline__ int
-metal_interrupt_set_priority(struct metal_interrupt *controller, int id,
```
nategraff-sifive: It looks like the `get_priority` and `set_priority` methods are getting deleted and not replaced?

zongbox: Yes, the first patch in this PR removes four APIs, because these four APIs are defined twice in the interrupt.c.

metal/interrupt.h
zongbox: OK, it is good to me. The relevant bit is 0 upon success, otherwise, it is 1.

zongbox:
Push the modification about changing the return type of affinity APIs.
This PR depends on freedom-devicetree-tools #150

zongbox:
@nategraff-sifive @bsousi5 ping. I had pushed the modification, please help to review that, thanks a lot. :)

PR #171, opened 2019-08-27
Add QSPI mode and update spi driver.

reclusejack:
QSPI_ALL
      _                                                        _
CS     |______________________________________________________|
         _   _   _   _   _   _   _   _   _   _   _   _   _   _
SCL    _| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_

Phase ---cmd--|-addr--|-dummy-|-data----------------------------
DQ0     0   4 | 0   4 | 0   4 | 0   4   8  12...
              |       |       |
DQ1     1   5 | 1   5 | 1   5 | 1   5   9  13...
              |       |       |
DQ2     2   6 | 2   6 | 2   6 | 2   6  10  14...
              |       |       |
DQ3     3   7 | 3   7 | 3   7 | 3   7  11  15...

QSPI_ADDR_DATA
      _                                                        _
CS     |______________________________________________________|
         _   _   _   _   _   _   _   _   _   _   _   _   _   _
SCL    _| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_

Phase  ------cmd------|-addr--|-dummy-|-data--------------------
DQ0     0   1   2   3 | 0   4 | 0   4 | 0   4   8  12...
                      |       |       |
DQ1                   | 1   5 | 1   5 | 1   5   9  13...
                      |       |       |
DQ2                   | 2   6 | 2   6 | 2   6  10  14...
                      |       |       |
DQ3                   | 3   7 | 3   7 | 3   7  11  15...

QSPI_DATA_ONLY
      _                                                        _
CS     |______________________________________________________|
         _   _   _   _   _   _   _   _   _   _   _   _   _   _
SCL    _| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_| |_

Phase  ------cmd------|------addr-----|-----dummy-----|-data----
DQ0     0   1   2   3 | 0   1   2   3 | 0   1   2   3 | 0   4
                      |               |               |
DQ1                   |               |               | 1   5
                      |               |               |
DQ2                   |               |               | 2   6
                      |               |               |
DQ3                   |               |               | 3   7

### DIFF
--- a/metal/spi.h
+++ b/metal/spi.h
@@ -21,6 +21,18 @@ struct metal_spi_config {
     unsigned int cs_active_high : 1;
     /*! @brief The chip select ID to activate for the SPI transfer */
     unsigned int csid;
+    /*! @brief The spi command frame number (cycles = num * frame_len) */
+    unsigned int cmd_num;
+    /*! @brief The spi address frame number */
+    unsigned int addr_num;
+    /*! @brief The spi dummy frame number */
+    unsigned int dummy_num;
+    /*! @brief The Dual/Quad spi mode selection.*/
+    enum {
+        MULTI_WIRE_ALL,
+        MULTI_WIRE_DATA_ONLY,
+        MULTI_WIRE_ADDR_DATA
+    } multi_wire;
 };
 
 struct metal_spi_vtable {
--- a/src/drivers/sifive_spi0.c
+++ b/src/drivers/sifive_spi0.c
@@ -63,10 +63,16 @@ static int configure_spi(struct __metal_driver_sifive_spi0 *spi,
         METAL_SPI_REGW(METAL_SIFIVE_SPI0_FMT) |= METAL_SPI_PROTO_SINGLE;
         break;
     case METAL_SPI_DUAL:
-        METAL_SPI_REGW(METAL_SIFIVE_SPI0_FMT) |= METAL_SPI_PROTO_DUAL;
+        if (config->multi_wire == MULTI_WIRE_ALL)
+            METAL_SPI_REGW(METAL_SIFIVE_SPI0_FMT) |= METAL_SPI_PROTO_DUAL;
+        else
+            METAL_SPI_REGW(METAL_SIFIVE_SPI0_FMT) |= METAL_SPI_PROTO_SINGLE;
         break;
     case METAL_SPI_QUAD:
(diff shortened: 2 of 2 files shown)


metal/spi.h
```diff
@@ -21,6 +21,14 @@ struct metal_spi_config {
...
     /*! @brief The chip select ID to activate for the SPI transfer */
     unsigned int csid;
+    /*! @brief The spi command frame number (cycles = num * frame_len) */
+    unsigned int cmd_num;
+    /*! @brief The spi address frame number */
+    unsigned int addr_num;
+    /*! @brief The spi dummy frame number */
+    unsigned int dummy_num;
+    /*! @brief The Dual/Quad spi mode selection.*/
+    enum { QSPI_ALL, QSPI_DATA_ONLY, QSPI_ADDR_DATA } qspi_mode;
```
nategraff-sifive: If this applies to both dual and quad-SPI, we should probably not name it "qspi_mode".

src/drivers/sifive_spi0.c
```diff
@@ -131,13 +137,36 @@ static int configure_spi(struct __metal_driver_sifive_spi0 *spi,
     return 0;
 }
 
+static void spi_mode_switch(struct __metal_driver_sifive_spi0 *spi,
+                            struct metal_spi_config *config,
+                            unsigned int qspi_mode) {
+    long control_base =
+        __metal_driver_sifive_spi0_control_base((struct metal_spi *)spi);
+
+    if (config->qspi_mode == qspi_mode) {
```
nategraff-sifive: Why is this check here? If they have to be equal, why not just use the one already in `config`?

metal/spi.h
reclusejack: Any suggestion for this variable name? How about "multi_wire" or "multi_wire_mode"?

src/drivers/sifive_spi0.c
reclusejack: That' my mistake, I should not use the same name for the two variables.
I've renamed these two variables. 
First one is used for SPI configuration, the second is for tagging the data transfer stage.
The function spi_mode_switch will be active either after sending out cmd or before sending data.

nategraff-sifive:
@reclusejack Can you issue a similar PR against v201908-branch? @dbarbi1 would like this in v19.08 as well.

PR #172, opened 2019-08-27
Retry if mcycle overflow when get cycle

kito-cheng:
It will get 0 without retry when mcycle overflow.


kito-cheng:
Re-write with do-while loop, the reason why first version using `goto` is it produce minimal diff :P

PR #191, opened 2019-10-04
Add macro & inline for get/set CSR

bsousi5:
This PR will add two types of APIs, an inline function and a macro for get and set CSR registers.
The inline function is not fully fill yet, but will get fill up with all supported CSR listed in https://github.com/riscv/riscv-opcodes. Its purpose is to provide some prior check before calling.
The macro is intended to be a raw form of CSR set/get.


kito-cheng:
This API will increasing 322 bytes on rv32ic or rv64ic with -Os for csr.c, and might be bigger once we fill up all supported CSR.

Just reminder the code size cost is high and might hard to optimized/reduce that in future.

nategraff-sifive:
Maybe the switch-case version should be removed from the API for code size in favor of just the macro.

PR #196, opened 2019-10-21
Always clear feature disable register

bsousi5:


reclusejack:
Hi @bsousi5 I never tried to enable chicken bit for cores other than 7-series. 
We need to check the regression result if it breaks anything~

src/drivers/riscv_plic0.c
```diff
@@ -136,12 +138,14 @@ void __metal_driver_riscv_plic0_init(struct metal_interrupt *controller) {
             /* Initialize ist parent controller, aka cpu_intc. */
             intc->vtable->interrupt_init(intc);
 
-            for (int i = 0; i < num_interrupts; i++) {
```
nategraff-sifive: Since we no longer use `num_interrupts`, we should probably remove the variable completely.

src/drivers/sifive_clic0.c
```diff
@@ -496,11 +498,15 @@ void __metal_driver_sifive_clic0_init(struct metal_interrupt *controller) {
         num_subinterrupts =
             __metal_driver_sifive_clic0_num_subinterrupts(controller);
         clic->metal_mtvt_table[0] = &__metal_clic0_handler;
-        for (int i = 1; i < num_subinterrupts; i++) {
```
nategraff-sifive: Similarly, remove `num_subinterrupts`.

mwachs5:
> Echoing @reclusejack, do we know what happens for non 7-series cores?

According to the spec, the right thing to do is always writing the register to 0 to enable all features

nategraff-sifive:
It probably makes sense to remove the` __metal_chicken_bit` from Freedom Devicetree Tools then, or deprecate it for a release by defining it everywhere before removing it the following release.

bsousi5:
> It probably makes sense to remove the` __metal_chicken_bit` from Freedom Devicetree Tools then, or deprecate it for a release by defining it everywhere before removing it the following release.

Agree. Will do that separately. But it should not gate this merge

PR #197, opened 2019-10-22
Simplify autotools configurations

nategraff-sifive:
The autotools configurations had a lot of cruft left over from early iterations of the library. This cleans up the build considerably, including:
- Removing the use of freedom-devicetree-tools within the Freedom Metal build itself
- Renaming the output archive files to libmetal, libmetal-gloss to match how they're consumed by Freedom E SDK
- Cleaning up various other legacy options which no longer affect the build

The result is a much cleaner build which is easier to follow and doesn't get lost in the noise of unused options and Make targets. Even if we go back to something like the old build system, I think it makes a lot more sense to reimplement that functionality in the context of new architectural decisions than to try and repurpose the old, bit-rotted build system.

This does require a few minor changes to Freedom E SDK.


bsousi5:
The clean up looks fine to me. However this will now make freedom-metal tightly couple with and  Freedom-e-sdk (which it is already doing) and depend on it to regress any changes.

Previously and originally freedom-metal was intended to be a stand-alone repo. I'll defer to @palmer-dabbelt  if he is OK with this.

PR #200, opened 2019-10-31
Picolibc

keith-packard:
Adds 'metal-pico' library which contains interfaces for picolibc to sit atop freedom-metal

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -157,6 +157,20 @@ libmetal_a_SOURCES = \
 	src/watchdog.c
 
 ########################################################
+# libmetal-pico
+########################################################
+
+lib_LIBRARIES += libmetal-pico.a
+
+libmetal_pico_a_SOURCES = \
+	gloss/crt0.S \
+	pico/iob.c \
+	gloss/sys_exit.c \
+	pico/times.c \
+	gloss/sys_sbrk.c \
+	pico/sysconf.c
+
+########################################################
 # libgloss
 ########################################################
 
--- a/gloss/sys_sbrk.c
+++ b/gloss/sys_sbrk.c
@@ -8,15 +8,20 @@
  * that's been allocated. */
 extern char metal_segment_heap_target_start;
 extern char metal_segment_heap_target_end;
-static char *brk = &metal_segment_heap_target_start;
+static char *__brk = &metal_segment_heap_target_start;
+
+#ifdef _PICOLIBC__
+#define _brk brk
+#define _sbrk sbrk
+#endif
 
 int _brk(void *addr) {
-    brk = addr;
+    __brk = addr;
     return 0;
 }
 
 char *_sbrk(ptrdiff_t incr) {
-    char *old = brk;
+    char *old = __brk;
 
     /* If __heap_size == 0, we can't allocate memory on the heap */
     if (&metal_segment_heap_target_start == &metal_segment_heap_target_end) {
@@ -24,10 +29,10 @@ char *_sbrk(ptrdiff_t incr) {
     }
 
     /* Don't move the break past the end of the heap */
-    if ((brk + incr) < &metal_segment_heap_target_end) {
-        brk += incr;
+    if ((__brk + incr) < &metal_segment_heap_target_end) {
+        __brk += incr;
     } else {
-        brk = &metal_segment_heap_target_end;
+        __brk = &metal_segment_heap_target_end;
         return (void *)-1;
     }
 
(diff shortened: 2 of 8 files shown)


pico/crt0.S
```diff
@@ -0,0 +1,250 @@
...
+  /* At this point we're in an environment that can execute C code.  The first
+   * thing to do is to make the callback to the parent environment if it's been
+   * requested to do so. */
+  beqz a2, 1f
+  jalr a2
+1:
+
+  /* Set TLS pointer */
+  la a0, __tls_base
+  call _set_tls
```
nategraff-sifive: It looks like this is the only thing that differs between gloss/crt0.S and pico/crt0.S, aside from a couple things which I'd actually recommend get put here. Can we find a way to put this in gloss/crt0.S so we can keep the same `_start` implementation?

pico/sbrk.c
```diff
@@ -0,0 +1,34 @@
...
+
+/* brk is handled entirely within the C library.  This limits METAL programs
+ * that use the C library to be disallowed from dynamically allocating memory
+ * without talking to the C library, but that sounds like a sane way to go
+ * about it.  Note that there is no error checking anywhere in this file, users
+ * will simply get the relevant error when actually trying to use the memory
+ * that's been allocated. */
+extern char metal_segment_heap_target_start;
+extern char metal_segment_heap_target_end;
+static char *_brk = &metal_segment_heap_target_start;
```
nategraff-sifive: The only difference I see between this an gloss/sys_sbrk.c is the underscore in `_brk`. Can we combine these into the same implementation?

Makefile.am
```diff
@@ -152,6 +152,20 @@ libmetal_a_SOURCES = \
 	src/vector.S \
 	src/watchdog.c
 
+########################################################
+# libpico_metal
+########################################################
+
+lib_LIBRARIES += libriscv__pico__metal.a
```
nategraff-sifive: Do we want to wrap this library in an option WITH_BUILTIN_LIBPICO_GLOSS or something similar, like we do for libgloss?

Also, is it worth matching the naming convention below more with a name like `libmetal-pico-gloss.a`?

nategraff-sifive: In your Freedom-E-SDK pull request you name it `libmetal-pico.a`, so we can match that name here instead of moving the file in that Makefile.

pico/crt0.S
keith-packard: Yup! Thanks for the suggestion. I've put the TLS initialization inlined right into gloss/crt0.S. I also shared sys_exit, but newlib-nano and picolibc need slightly different sbrk implementations (picolibc wants sbrk, newlib-nano wants _sbrk and _brk).

pico/sbrk.c
keith-packard: I think its a short enough piece of code that duplicating it will end up being less hassle than managing #ifdef's in the code? I've added the #ifdef's and pushed out the change; does that look reasonable?

Makefile.am
keith-packard: done

nategraff-sifive:
I'm happy with either version of sbrk, but the ifdef version might be slightly easier to maintain?

keith-packard:
> I'm happy with either version of sbrk, but the ifdef version might be slightly easier to maintain?

let's use the ifdef one then; we can always switch back :-)

keith-packard:
I've rebased and tested this with Nate's updated BSP generation code

gloss/sys_sbrk.c
```diff
@@ -8,26 +8,31 @@
...
-static char *brk = &metal_segment_heap_target_start;
+static char *__brk = &metal_segment_heap_target_start;
+
+#ifdef _PICOLIBC__
+#define _brk brk
+#define _sbrk sbrk
+#endif
 
 int _brk(void *addr) {
-    brk = addr;
```
bsousi5: what set brk now? is it newlib? how come there was no redefinition error previously?

Makefile.am
```diff
@@ -156,6 +156,20 @@ libmetal_a_SOURCES = \
 	src/vector.S \
 	src/watchdog.c
 
+########################################################
+# libmetal-pico
+########################################################
+
+lib_LIBRARIES += libmetail-pico.a
```
nategraff-sifive: Spelling `s/metail/metal/`

gloss/sys_sbrk.c
keith-packard: I had to rename the local 'brk' variable because picolibc uses that for the function newlib calls '_brk'. If you read the POSIX spec, 'brk' is the usual name; I'm not sure why newlib changed it.

PR #204, opened 2019-11-22
Create a ringbuffer implementation

nategraff-sifive:
This creates a ringbuffer implementation we can use as a default buffer for asynchronous peripheral APIs.

This implementation is specifically **not thread-safe** and is intended to be replaced by the thread-safe queue implementation of the API consumer's choice which is entitled to provide thread-safety in the context of the consumer's chosen kernel.

As long as I was in here, I thought I'd throw in some error codes from `errno.h`. Any thoughts on that decision instead of the current "return random non-zero numbers on failure" of other APIs?

### DIFF
--- a/metal/ringbuf.h
+++ b/metal/ringbuf.h
@@ -85,7 +85,7 @@ size_t metal_ringbuf_num_items(struct metal_ringbuf *rb);
  *
  * @param rb The ring buffer
  * @param val The value to append
- * @return 0 if the value is added, nonzero if no space remains
+ * @return 0 if the value is added, -ENOMEM if no space remains
  */
 int metal_ringbuf_put(struct metal_ringbuf *rb, const void *val);
 
@@ -100,7 +100,7 @@ int metal_ringbuf_put(struct metal_ringbuf *rb, const void *val);
  *
  * @param rb The ring buffer
  * @param val Stores the popped value
- * @return 0 if a value is popped, nonzero if the buffer is empty
+ * @return 0 if a value is popped, -ENODATA if the buffer is empty
  */
 int metal_ringbuf_get(struct metal_ringbuf *rb, void *val);
 
@@ -113,7 +113,7 @@ int metal_ringbuf_get(struct metal_ringbuf *rb, void *val);
  *
  * @param rb The ring buffer
  * @param val Stores the popped value
- * @return 0 if a value is peeked, nonzero if the buffer is empty
+ * @return 0 if a value is peeked, -ENODATA if the buffer is empty
  */
 int metal_ringbuf_peek(struct metal_ringbuf *rb, void *val);
 
--- a/metal/ringbuf.h
+++ b/metal/ringbuf.h
@@ -12,23 +12,27 @@
  * @brief The handle for a ring buffer
  */
 struct metal_ringbuf {
-    /* _buf points at the base of the ring buffer memory */
-    void *_buf;
-    /* _start points at the ring buffer location containing the first value to
-     * be read by metal_ringbuf_get */
-    void *_start;
-    /* _end points at the first empty location after the buffer contents */
-    void *_end;
-    /* _capacity is the number of items the ring buffer can hold */
-    size_t _capacity;
-    /* _item_size is the size of each element in the ring buffer memory */
-    size_t _item_size;
-    /* _empty is 0 when the buffer is empty, 1 otherwise. This is used to
-     * differentiate
-     * the full condition from the empty condition, for which both cases have
-     * _start == _end */
-    bool _empty;
+    /* _read is the index into _buf to begin reading from the buffer */
+    uint32_t _read;
+    /* _count is the number of elements in the buffer */
+    uint32_t _count;
+    /* _capacity is the total number of bytes in the buffer */
+    uint32_t _capacity;
+    /* A Flexible Array Member for holding the contents of the ringbuffer,
+     * requires C >= C99 and must be the last element in the struct */
+    uint8_t _buf[];
 };
+
+/*!
+ * @def METAL_RINGBUF_SIZEOF
+ * @brief Get the required size of the memory to hold a ringbuf of N bytes
+ *
+ * When allocating memory for use in `metal_ringbuf_create()`, users should use
+ * this macro to determine the amount of memory to allocate for a ring buffer
+ * capacity of `capacity` bytes.
+ */
+#define METAL_RINGBUF_SIZEOF(capacity)                                         \
+    (sizeof(struct metal_ringbuf) + (capacity))
 
 /*!
  * @def METAL_RINGBUF_DECLARE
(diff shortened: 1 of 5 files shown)


metal/ringbuf.h
```diff
@@ -0,0 +1,119 @@
...
+#define METAL__RINGBUF_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/*!
+ * @brief The handle for a ring buffer
+ */
+struct metal_ringbuf {
```
bsousi5: We should also have another structure for item, that have a pointer and size of the item, which metal_ringbuf will include. Plus API to create this item, return item, its value, item size.

src/ringbuf.c
```diff
@@ -0,0 +1,144 @@
...
+#include <errno.h>
+#include <string.h>
+
+#include <metal/ringbuf.h>
+
+struct metal_ringbuf metal_ringbuf_create(void *buf, size_t capacity,
+                                          size_t item_size) {
+    struct metal_ringbuf rb;
+
+    rb._buf = buf;
```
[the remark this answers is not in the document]
nategraff-sifive: Noted, I've changed the API to add the NULL check.

src/ringbuf.c
```diff
@@ -0,0 +1,144 @@
...
+
+    if (rb->_start == rb->_end) {
+        /* mark the buffer as now empty */
+        rb->_empty = true;
+    }
+
+    return 0;
+}
+
+int metal_ringbuf_peek(struct metal_ringbuf *rb, void *val) {
```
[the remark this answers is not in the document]
nategraff-sifive: That's intentional. I'm trying to make this compatible with queues in FreeRTOS and Zephyr RTOS, which do leave this burden up to the caller.

metal/ringbuf.h
nategraff-sifive: That's a whole `sizeof(size_t)` per item additional overhead for a value which is constant for all items in the ring buffer. [FreeRTOS's Queues only require (capacity * item size) bytes of storage.](https://freertos.org/xQueueCreateStatic.html)

metal/ringbuf.h
```diff
@@ -0,0 +1,120 @@
...
+    void *_end;
+    /* _capacity is the number of items the ring buffer can hold */
+    size_t _capacity;
+    /* _item_size is the size of each element in the ring buffer memory */
+    size_t _item_size;
+    /* _empty is 0 when the buffer is empty, 1 otherwise. This is used to
+     * differentiate
+     * the full condition from the empty condition, for which both cases have
+     * _start == _end */
+    bool _empty;
```
keith-packard: You can avoid the separate _empty boolean by using a 'count' value that tracks how much data is in the ring instead of  '_start'.
If you do choose to do this, I'd suggest using an integer for _end too. Oh, names are hard; _end and _count are not exactly clear. I've used 'read' and 'count' to indicate the place to start reading and the count of data in the buffer, but you might have better ideas.

metal/ringbuf.h
```diff
@@ -0,0 +1,120 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/*!
+ * @brief The handle for a ring buffer
+ */
+struct metal_ringbuf {
+    /* _buf points at the base of the ring buffer memory */
+    void *_buf;
```
keith-packard: Consider how you might remove this from the struct by placing the data array at the end instead of in a separate variable. That will take some tricky CPP bits, but you would save a bunch of memory references fetching the pointer and save the space of the pointer in RAM.

metal/ringbuf.h
```diff
@@ -0,0 +1,120 @@
...
+ * This can be used in place of METAL_RINGBUF_DECLARE to create a ring buffer at
+ * runtime
+ *
+ * @param rb A pointer to a struct metal_ringbuf to be initialized
+ * @param buf The memory for the buffer
+ * @param capacity The number of items the buffer can hold
+ * @param item_size the size of each item in bytes
+ * @return 0 if successfully created, -EINVAL if rb or buf are NULL
+ */
+int metal_ringbuf_create(struct metal_ringbuf *rb, void *buf, size_t capacity,
```
keith-packard: Would it make sense for this function to just call malloc? If the space is not dynamically allocated, then I don't think this API will be better than the macro interface above?

metal/ringbuf.h
```diff
@@ -0,0 +1,120 @@
...
+    void *_buf;
+    /* _start points at the ring buffer location containing the first value to
+     * be read by metal_ringbuf_get */
+    void *_start;
+    /* _end points at the first empty location after the buffer contents */
+    void *_end;
+    /* _capacity is the number of items the ring buffer can hold */
+    size_t _capacity;
+    /* _item_size is the size of each element in the ring buffer memory */
+    size_t _item_size;
```
keith-packard: This does make the system more general, but I suspect the dominant use will be to buffer bytes. Would it be better to offer APIs that insert/remove multiple bytes and remove this field? It would also ensure that the amount of memory read/written by the queue/dequeue functions was part of that API, rather than being implicit from the configuration of the ring.

metal/ringbuf.h
nategraff-sifive: I thought I'd let the user decide how they want to allocate the memory for the buffer. `malloc` remains an option, but this way the user could put the buffer on the stack, or manually create a global with the section of their choice so that the linker places it in a specific memory region, etc. The macro interface then is just a shorthand that means "put it in `.data`".

metal/ringbuf.h
nategraff-sifive: Oh, you mean make the buffer bigger and store the bookkeeping at the beginning? Hmmmm

nategraff-sifive: We're using C99, so this should be achievable with a Flexible Array Member. If there's any C++ trickery needed I don't see it yet.

nategraff-sifive:
@keith-packard I just pushed a commit which makes the default data type of the buffer a byte and uses a Flexible Array Member to co-locate the buffer and bookkeeping in a single struct. I do like these changes, but I wonder if you have any thoughts about how this affects being able to swap out this implementation for an RTOS's Queue.

keith-packard:
> @keith-packard I just pushed a commit which makes the default data type of the buffer a byte and uses a Flexible Array Member to co-locate the buffer and bookkeeping in a single struct. I do like these changes, but I wonder if you have any thoughts about how this affects being able to swap out this implementation for an RTOS's Queue.

If the low-level drivers don't directly interact with the ring buffer, or interact through a simple macro interface, then I think it will be easy to replace that with calls into the RTOS implementation. This again encourages us to think of freedom-metal as a pile of source code which is compiled into applications rather than a library of pre-compiled bits :-)

I guess my last question is whether we actually need to make the ring buffer able to have different sizes for different uses in the same application. If the size of the ring buffer was a compile time constant, then you could use a fixed size array rather than a flexible sized array. This would make static allocation easier as you wouldn't need to do a bunch of magic to ensure that the array allocation followed the struct. Hrm. C sucks sometimes.

PR #205, opened 2019-12-10
Build with segger libc v1911

issuehsu:
This PR is to adding SEGGER libc support into freedom-e-sdk and freedom-metal


issuehsu:
> Instead of reimplementing crt0.S, sys_exit, etc. in `segger/*`, the Segger-based library should just pull in the existing files from `gloss/*` so we don't have to maintain two implementations in sync.

Will be fixed.

issuehsu:
I will close this PR and create another one for newer updates

PR #206, opened 2019-12-11
Add support for Sifive PWM.

NandkumarJoshi:
Added files for PWM driver. Minor updates to I2C. IO pin mux handling is changed, will affect pin mux values in dts. Current IO mux handling doesn't allow to 'set' a bit in 'IOF_SEL' register.


NandkumarJoshi:
will raise a related PR in E-SDK, for updates in DTS, addition of an example.

PR #211, opened 2019-12-23
ESAT-213 fix

NandkumarJoshi:
SP 64bit alignment is now checked in crt0.S

### DIFF
--- a/gloss/crt0.S
+++ b/gloss/crt0.S
@@ -44,22 +44,7 @@ _start:
   la gp, __global_pointer$
 .option pop
 
-  /* The METAL is designed for a bare-metal environment and therefor is expected
-   * to define its own stack pointer.  We also align the stack pointer here
-   * because the only RISC-V ABI that's currently defined mandates 16-byte
-   * stack alignment. */
-  la sp, _sp
-
-  /* Increment by hartid number of stack sizes */
-  li t0, 0
-  la t1, __stack_size
-1:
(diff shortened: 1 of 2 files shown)


gloss/crt0.S
```diff
@@ -44,12 +44,19 @@ _start:
...
 
-  /* The METAL is designed for a bare-metal environment and therefor is expected
+  /* The METAL is designed for a bare-metal environment and therefore is expected
    * to define its own stack pointer.  We also align the stack pointer here
    * because the only RISC-V ABI that's currently defined mandates 16-byte
    * stack alignment. */
+  bne sp, zero, 1f
   la sp, _sp
+1:
+  andi t0, sp, 7
```
kito-cheng: Stack alignment is 16-byte, so I think should be 15 not 7?

gloss/crt0.S
```diff
@@ -59,7 +66,7 @@ _start:
   addi t0, t0, 1
   j 1b
 1:
-  andi sp, sp, -16
+  addi sp, sp, -16
```
kito-cheng: Why `addi` instead of `andi`? `andi sp, sp, -16` align the stack to 16 bytes, but `addi sp, sp, -16` seems like just allocate 16 bytes stack?

bsousi5:
I assume, you will still make changes to entry.S. As it, for MC the stack may not be 16bytes aligned. Only the last hart stack will.

PR #215, opened 2019-12-26
Fix warnings as error issues when following options applied

issuehsu:
Fix warning as error issues when following options applied:

-std=c99 -Wall -Wextra -Werror -Wpedantic -Werror -Wno-unused-parameter

### DIFF
--- a/src/i2c.c
+++ b/src/i2c.c
@@ -19,7 +19,7 @@ extern inline int metal_i2c_get_baud_rate(struct metal_i2c *i2c);
 extern inline int metal_i2c_set_baud_rate(struct metal_i2c *i2c, int baud_rate);
 
 struct metal_i2c *metal_i2c_get_device(int device_num) {
-    if (device_num >= __METAL_DT_MAX_I2CS) {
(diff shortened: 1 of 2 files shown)


src/i2c.c
```diff
@@ -19,7 +19,7 @@ extern inline int metal_i2c_get_baud_rate(struct metal_i2c *i2c);
 extern inline int metal_i2c_set_baud_rate(struct metal_i2c *i2c, int baud_rate);
 
 struct metal_i2c *metal_i2c_get_device(int device_num) {
-    if (device_num >= __METAL_DT_MAX_I2CS) {
+    if (device_num >= __METAL_DT_MAX_I2CS || device_num < 0) {
```
bsousi5: Can use the following instead,
#if __METAL_DT_MAX_I2CS > 0
    if (device_num < __METAL_DT_MAX_I2CS) {
        return (struct metal_i2c *)__metal_i2c_table[device_num];
    }
#endif

Thanks for doing this. We missed when it was added.
    return NULL;

PR #224, opened 2020-01-08
Wait 100usec before testing FE310 PLL for lock

### LINKED ISSUE #213: configure_pll() might be in a rush to test PLL_LOCK
Around line 293 of src/drivers/sifive_fe310-g000_pll.c:

```
    /* Wait for PLL to lock */
    while((__METAL_ACCESS_ONCE(pllcfg) & PLL_LOCK) == 0) ;
```

Section 6.5 of the FE310-G002 Manual 1p4 says:

> The PLL provides a lock signal which is set when the PLL has achieved lock, and which can be
> read from the most-significant bit of the pllcfg register. The PLL requires up to 100 μs to
> regain lock once enabled, and the lock signal will not necessarily be stable during this initial lock
> period so should only be interrogated after this period.

So this code seems to be a little aggressive, not waiting 100uS before entering the wait loop. I have not confirmed why yet, but a test program that configures the PLL to 320MHz infrequently hangs apparently when setting the PLL. Hardware is SparkFun Red-V with R8 corrected to 100 ohms (power supply filter for PLLVDD).

I appreciate a 100uS delay might be hard to arrange here in Metal given it seems to depend on other hardware configuration (LFROSC?) or CPU clock rate.

nategraff-sifive:
The FE310 PLL needs 100 usec to stabilize before testing the lock signal. This solution assumes that mtime is driven at 32768 Hz for all targets with the FE310 PLL.

Should this be backported to v19.08?

Fixes #213

### DIFF
--- a/src/drivers/sifive_fe310-g000_pll.c
+++ b/src/drivers/sifive_fe310-g000_pll.c
@@ -10,6 +10,7 @@
 
 #include <metal/init.h>
 #include <metal/machine.h>
+#include <metal/machine/platform.h>
 
 #include <metal/drivers/sifive_fe310-g000_pll.h>
 #include <stdlib.h>
@@ -270,6 +271,19 @@ static int find_closest_config(long ref_hz, long rate) {
     return closest_index;
 }
 
+/* The PLL needs 100 usec to stabilize before we test PLL_LOCK. Since LFROSC
+ * on all targets with the FE310-G000 PLL runs at 32768 Hz, we need to wait
+ * at least
+ *
+ *   ceil(100 usec * 32768 ticks/sec * 1 sec / 1000000 usec) = 4 ticks
+ *
+ * of mtime before we test PLL_LOCK.
+ *
+ * TODO: Determine the mtime timebase at compile or runtime and use that
+ * here.
+ */
+#define PLL_LOCK_WAIT_TICKS 4
+
 /* Configure the PLL and wait for it to lock */
 static void configure_pll(__metal_io_u32 *pllcfg, __metal_io_u32 *plloutdiv,
                           const struct pll_config_t *config) {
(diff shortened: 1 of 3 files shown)


src/drivers/sifive_fe310-g000_pll.c
```diff
@@ -270,6 +272,11 @@ static int find_closest_config(long ref_hz, long rate) {
     return closest_index;
 }
 
+int __metal_driver_riscv_clint0_command_request(
+    struct metal_interrupt *controller, int command, void *data);
+int __metal_driver_sifive_clic0_command_request(
+    struct metal_interrupt *controller, int command, void *data);
+
```
keith-packard: These declarations should be in a header file that is included both here and by the file defining the functions so that these declarations can be typechecked against the definitions.

src/drivers/sifive_fe310-g000_pll.c
```diff
@@ -297,6 +304,40 @@ static void configure_pll(__metal_io_u32 *pllcfg, __metal_io_u32 *plloutdiv,
 
     __METAL_ACCESS_ONCE(pllcfg) &= ~(PLL_BYPASS);
 
+    /* The PLL needs 100 usec to stabilize before we test PLL_LOCK. Since LFROSC
+     * on all targets with the FE310-G000 PLL runs at 32768 Hz, we need to wait
+     * at least
+     *
+     *   ceil(32768 ticks * 100 usec / 1000000 usec) = 4 ticks
```
keith-packard: I like using dimensional analysis to validate conversions like this; let's be careful and use the right dimensions:

32768 ticks/sec * 1 sec / 1e6 µsec * 100 µsec = 3.2768 ticks/100µsec

src/drivers/sifive_fe310-g000_pll.c
```diff
@@ -297,6 +304,40 @@ static void configure_pll(__metal_io_u32 *pllcfg, __metal_io_u32 *plloutdiv,
...
+     * of mtime before we test PLL_LOCK.
+     *
+     * TODO: Determine the mtime timebase at compile or runtime and use that
+     * here.
+     */
+#ifdef __METAL_DT_RISCV_CLINT0_HANDLE
+    unsigned long long mtime, mtime_end;
+    __metal_driver_riscv_clint0_command_request(__METAL_DT_RISCV_CLINT0_HANDLE,
+                                                METAL_TIMER_MTIME_GET, &mtime);
+    mtime_end = mtime + 4;
```
keith-packard: Let's use a #define near the comment above to set the interval so it's clear what this constant '4' means

src/drivers/sifive_fe310-g000_pll.c
```diff
@@ -297,6 +304,40 @@ static void configure_pll(__metal_io_u32 *pllcfg, __metal_io_u32 *plloutdiv,
...
+    while (mtime_end > mtime) {
```
keith-packard: I think you're off by one here -- if we read the first time near the end of the tick, then we may only wait for 3+ε ticks instead of the required 4.

Something like while (mtime <= mtime_end) should do the trick

src/drivers/sifive_fe310-g000_pll.c
nategraff-sifive: I just combined the `32768 ticks/sec * 1 sec/1e6 usec` into a single `32768 ticks / 1e6 usec`, since I usually do that shorthand in my own dimensional analysis, but I can make this more explicit.

src/drivers/sifive_fe310-g000_pll.c
nategraff-sifive: You're right, thanks!

src/drivers/sifive_fe310-g000_pll.c
nategraff-sifive: Moved to the header files for the CLIC/CLINT drivers.

PR #227, opened 2020-01-10
cleanup wake code, inline autoconf_wrapper script

albertchen-sifive:
inline `autoconf_wrapper` so they don't get out of sync. also improve error reporting and make `Job` available in outputs


nategraff-sifive:
@albertchen-sifive Can you rebase this against latest master?

PR #228, opened 2020-01-12
Memory address can exceed 32bits

bsousi5:


kito-cheng:
Is this patch trying to handle more than 32 bits addressing space in RV32? If so, I think that mean no uintptr_t should used in whole freedom-metal?

If this patch is just for RV64, I assume uintptr_t is already 64 bits?

bsousi5:
> Is this patch trying to handle more than 32 bits addressing space in RV32? If so, I think that mean no uintptr_t should used in whole freedom-metal?
> 
> If this patch is just for RV64, I assume uintptr_t is already 64 bits?

This is not to fix RV64. It is for a RV32, where its base address + size can exceed 32bits.
I agree for handling more than 32bits addressing space in RV32, we will need to large code more and can no longer rely on uintptr_t. That would be a complete review and changes agrees freedom-metal.

kito-cheng:
> This is not to fix RV64. It is for a RV32, where its base address + size can exceed 32bits.

Curious about how to access those data which exceed 32 bits addressing space in RV32, there is some CSR like segment register in x86?

PR #231, opened 2020-03-10
Fix synchronization - continue to execute on the core selected on the dts file

e-puerto:


nategraff-sifive:
Is there a reason the synchronization is hart-sensitive? It shouldn't really matter which hart coordinates the synchronization.

e-puerto:
will not be added

PR #235, opened 2020-03-25
Interrupt Preemption support (CLIC) for nested interrupts

bsousi5:
Provide two new APIs to set and get interrupt preemptive level, primarily for CLIC.

Also update cache flush API to skip per line flushing where Cores does not support.


src/cache.c
```diff
@@ -115,76 +115,68 @@ int metal_icache_l1_available(int hartid) {
...
  * state machine in L1 Data Cache (D$) with funct3=0, (for core with data
  * caches) It is an I type: .insn i opcode, func3, rd, rs1, simm12(signed
- * immediate 12bs) 31     28 27    24 23    20 19     16 15   12 11     8 7 4 3
- * 0
+ * immediate 12bs)
+ * 31     28 27    24 23    20 19    16 15    12 11     8 7      4 3       0
  * |--------|--------|--------|--------|--------|--------|--------|--------|
  * +-------------+------------+----------+------+--------+-----------------+
  * |sign immediate12b (simm12)|   rs1    | func3|    rd  |      opcode     |
  * |-1-1-1-1 -1-1-0-0 -0-0-0-0|-x-x-x-x-x|0-0-0-|-0-0-0-0|-0-1-1-1 -0-0-1-1|
```
e-puerto: -1-1-1-1 -1-1-0-0 -0-0-0-0 is for -0x3F if the value is -0X3E it should be -1-1-1-1 -1-1-0-0 -0-0-0-1

src/cache.c
```diff
@@ -115,76 +115,68 @@ int metal_icache_l1_available(int hartid) {
...
-        __asm__ __volatile__(".insn i 0x73, 0, x0, x0, -0x3F" : :);
-        __asm__ __volatile__("fence.i"); // FENCE
+        if (address) {
+            // Using ‘.insn’ pseudo directive:
+            //       '.insn i opcode, func3, rd, rs1, simm12'
+            __asm__ __volatile__(".insn i 0x73, 0, x0, %0, -0x3E"
+                                 :
+                                 : "r"(address));
+        } else {
+            __asm__ __volatile__(".word 0xfc200073" : : : "memory");
```
e-puerto: 0xfc200073 is not equivalent to .insn i 0x73, 0, x0, %0, -0x3E when %0 = 0, the value should be 0xfc100073

src/cache.c
```diff
@@ -115,76 +115,68 @@ int metal_icache_l1_available(int hartid) {
...
  *            the virtual address in integer register rs1.
  */
 void metal_dcache_l1_flush(int hartid, uintptr_t address) {
     if (metal_dcache_l1_available(hartid)) {
-        // Using ‘.insn’ pseudo directive: '.insn i opcode, func3, rd, rs1,
-        // simm12'
-        __asm__ __volatile__(".insn i 0x73, 0, x0, %0, -0x40" : : "r"(address));
-        __asm__ __volatile__("fence.i"); // FENCE
+        if (address) {
+            uintptr_t ms1, ms2;
```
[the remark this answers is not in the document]
bsousi5: Prefer to leave it to the compiler to pick the one it need.

metal/interrupt.h
```diff
@@ -319,6 +323,35 @@ metal_interrupt_get_priority(struct metal_interrupt *controller, int id) {
...
+/*!
+ * @brief Get an interrupt preemptive level
+ * @param controller The handle for the interrupt controller
+ * @param id The interrupt ID to enable
+ * @return The interrupt level
+ */
+__inline__ unsigned int
+metal_interrupt_get_preemptive_level(struct metal_interrupt *controller,
+                                     int id) {
+    return controller->vtable->interrupt_get_preemptive_level(controller, id);
```
nategraff-sifive: I see that the CLIC vtable is getting this, but the CLINT isn't? There should be a NULL check to make sure that calls to this public API don't dereference a NULL pointer for CLINT instances.

src/cache.c
bsousi5: Oh, yes will do. Thanks for pointing out.

metal/interrupt.h
bsousi5: Oh, will add a stub function for Clint. Thanks

PR #236, opened 2020-03-27
Stop mangling newlines

knightsifive:
Currently, when `__METAL_DT_STDOUT_UART_HANDLE` is defined, `metal_tty_putc` mangles newlines by replacing `\n` with `\r\n`, ultimately resulting in non-conformant Newlib standard output routines. This pull request removes this mangling. 

This behavior is particularly undesirable when standard output (over UART) is used to transmit binary data: a `0x0D` is quietly injected before every `0x0A` in the byte-stream. 

This non-conformant behavior might be desirable on Windows systems, or for transmission using the HTTP protocol, where `\r\n` is interpreted as a newline. However, the behavior is undesirable on Linux and Mac OS (v10+) systems, which use only `\n`, or Mac OS (< v10) systems, which use only `\r`. It should be the user's responsibility to emit the correct newline sequence for their (host) platform. 

Note that "sifive/example-hello", which features only `\n`, may now fail to display the "expected" result on certain Windows terminal emulators. Most terminal emulators provide a mechanism for dealing with this: for example, in PuTTY:

```
Some servers only send LF, and expect the terminal to move the cursor over to the left automatically. If you come across a server that does this, you will see a stepped effect on the screen, like this:

First line of text
                  Second line
                             Third line

If this happens to you, try enabling the ‘Implicit CR in every LF’ option, and things might go back to normal:

First line of text
Second line
Third line
```


knightsifive:
This PR not only impacts Newlib's IO routines (as explained above), but also @keith-packard's PicoLibc routines, via
https://github.com/sifive/freedom-metal/blob/c0fe92120d8ad6a321999ec8fe006636da742ef8/pico/iob.c#L28

nategraff-sifive:
Bumping this thread. @keith-packard I remember you had some analysis for why we need to be able to have it both ways somehow?

knightsifive:
The only motivation I can imagine for this "feature" is to improve Freedom-E-SDK OOBE for a subset of Windows users. Modern Windows users have switched from PuTTY to WSL Bash, which understands `\n`. 

A saner approach would be an `#ifdef WINDOWS_HOST` in example-hello and friends that uses `\r\n` instead of `\n`. A slightly less sane approach --- but still better than the status quo --- is to similarly `#ifdef`-wrap the Freedom-Metal code deleted by this PR. In either case, the user could then add `-DWINDOWS_HOST` to Freedom-E-SDK's `CCFLAGS`.

knightsifive:
@bsousi5: Note that @RajeshVaradharajan's `putc` doesn't mangle newlines:
https://github.com/sifive/freedom-metal/blob/f7a3c9cbb793f66b87a7dca01dcee981a78a7c6d/src/drivers/sifive_simuart0.c

bsousi5:
> @bsousi5: Note that @RajeshVaradharajan's `putc` doesn't mangle newlines:
> https://github.com/sifive/freedom-metal/blob/f7a3c9cbb793f66b87a7dca01dcee981a78a7c6d/src/drivers/sifive_simuart0.c
> 
> EDIT: on second thought, maybe it does. Sorry if I'm wrong. In any case, I still think this PR warrants attention.

@RajeshVaradharajan PR#261 does not make any change to metal_tty_putc(). Only adding a sim-uart driver.

mwachs5:
as the originator of this awfulness I approve this PR based on our slack conversation

bsousi5:
> as the originator of this awfulness I approve this PR based on our slack conversation

Will merged this once I resolved the Windows issue with @sifivekevin for Freedom Studio.

PR #237, opened 2020-03-27
Add driver for SiFive Bus Error Unit

nategraff-sifive:
I'm opening a draft pull request for this API to seek feedback before I go to the trouble of implementing it.

I'm ditching the vtable pattern for this driver because I don't have any reason to believe that this API generalizes. The `metal_buserror` prefix will be public API surface and always refer to the `sifive,buserror0` device or its successors. My only question is if we should keep the code in the `drivers` folder.

I referenced Chapter 7 of the [u54mc Manual](https://sifive.cdn.prismic.io/sifive/86e05812-e9cd-4553-bfef-c7e715088055_sifive_coreip_U54MC_AXI4_rtl_v19_08p2p0_release_manual.pdf) when writing this API.


metal/drivers/sifive_buserror0.h
```diff
@@ -0,0 +1,121 @@
...
+
+/*!
+ * @brief Enable or disable the hart-local interrupt
+ *
+ * @param beu The bus error unit handle
+ * @param event The error event which would trigger the interrupt
+ * @param enabled True if the interrupt should be enabled
+ * @return 0 upon success
+ */
+int metal_buserror_set_local_interrupt(struct metal_buserror *beu, metal_buserror_event_t event, bool enabled);
```
bsousi5: The intention is the handle will have the hart-id?

metal/drivers/sifive_buserror0.h
```diff
@@ -0,0 +1,121 @@
...
+
+/*!
+ * @brief Get the error event which caused the most recent interrupt
+ *
+ * This method should be called from within the interrupt handler for the bus error unit interrupt
+ *
+ * @param beu The bus error unit handle
+ * @return The event which caused the interrupt
+ */
+metal_buserror_event_t metal_buserror_get_cause(struct metal_buserror *beu);
```
bsousi5: when happen if it call from outside ISR? will it be safe or simply not guarantee? or not allowed?

metal/drivers/sifive_buserror0.h
```diff
@@ -0,0 +1,121 @@
...
+/*!
+ * @brief Clear the cause register for the bus error unit
+ *
+ * This method should be called from within the interrupt handler for the bus error unit to un-latch
+ * the cause register for the next event
+ *
+ * @param beu The bus error unit handle
+ * @return 0 upon success
+ */
+int metal_buserror_clear_cause(struct metal_buserror *beu);
```
bsousi5: same as above

metal/drivers/sifive_buserror0.h
nategraff-sifive: Yeah, the bus error unit handle will know which hart ID it's associated with.

metal/drivers/sifive_buserror0.h
nategraff-sifive: The `cause` register is set to 0, indicating no error, when no event has occurred, so it should be valid to call this outside an ISR. It'll just tell you nothing has happened.

metal/drivers/sifive_buserror0.h
nategraff-sifive: Writing 0 to the cause register should be safe outside of an ISR.

metal/drivers/sifive_buserror0.h
nategraff-sifive: Sorry, I was confused. The BEU doesn't need to know its hart ID because this is the interrupt enable within the BEU. The user knows the hart ID because the BEU handle is retrieved from an individual CPU handle.

nategraff-sifive: Plus, the BEU interrupt is permanently enabled.

bsousi5: Are you planning to save the hart-id in the handle?

metal/drivers/sifive_buserror0.h
bsousi5: More concern when a non-isr thread calling this api, while beu isr is taken. Does the non-isr need to block interrupt before calling and unblock afterward?

metal/drivers/sifive_buserror0.h
nategraff-sifive: It won't need it.

metal/drivers/sifive_buserror0.h
nategraff-sifive: Outside of the ISR there should be no need to call this API.

nategraff-sifive:
Here's how I'm expecting to use the API: [example-buserror](https://github.com/sifive/example-buserror/blob/5a374e170dde18a94b5c52336516b813118ea8c3/buserror.c)

metal/drivers/sifive_buserror0.h
bsousi5: That's right the beu is already specific to a hart, and there is already per hart interrupt-controller/interrupt-table. I am good. Thanks

keith-packard:
I'd expect to see this API using masks to indicate a set of errors to handle in a particular way,
rather than a per-error-class API call. That might look like this:
    
    metal_buserror_set_event_enabled(beu, events);
    metal_buserror_set_platform_interrupt(beu, events);
    metal_buserror_set_local_interrupt(beu, events);
    metal_buserror_event_t metal_busserror_get_accrued(beu);
    metal_buserror_clear_event_acrrued(beu, events);

Define the metal_buserror_event_t as masks instead of values so that you
can easily | them together.

nategraff-sifive:
@keith-packard Yeah, better to accept an event mask. Thanks! I'll rework the API.

src/drivers/sifive_buserror0.c
```diff
@@ -0,0 +1,240 @@
...
+        return 1;
+    }
+    if (!(events & METAL_BUSERROR_EVENT_ANY)) {
+        return 2;
+    }
+
+    uint8_t *reg_enable = base + METAL_SIFIVE_BUSERROR0_ENABLE;
+
+    if (enabled) {
+        *reg_enable |= events;
```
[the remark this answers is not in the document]
nategraff-sifive: Oh, I forgot about that haha. I can reuse that, give me a minute.

nategraff-sifive:
> I don't see that we setup BEU as part of metal_init() when BEU is available. We would want to enable it by default.

How do we want it enabled? Setting the `enable` bits just causes events to be accrued, is that sufficient?

nategraff-sifive:
Branch updated to use `__METAL_ACCESS_ONCE` and a constructor has been added to enable all events by default.

PR #239, opened 2020-04-02
Initial stab 🗡

bsousi5:
Scrub memory based of what available in the lds files, ie
```
      PROVIDE( metal_itim_memory_start = ADDR(.itim) );
      PROVIDE( metal_itim_memory_end = ADDR(.itim) + SIZEOF(.itim) );
      PROVIDE( metal_dtim_memory_start = ADDR(.dtim) );
      PROVIDE( metal_dtim_memory_end = ADDR(.dtim) + SIZEOF(.dtim) );
      PROVIDE( metal_main_memory_start = ADDR(.ram) );
      PROVIDE( metal_main_memory_end = ADDR(.ram) + SIZEOF(.ram) );
```
And have scrub.S implement the __metal_before_start() which call metal_memory_scrub()
User/Customer have the option to override with their own if desire.

The intention is for metal_init() to enable the BEU and disable ECC suppression


src/scrub.S
```diff
@@ -0,0 +1,49 @@
...
+2:
+    ret
+
+/*
+ * Initialize all memories available to zero
+ * This must be call before seting up any stack(s)
+ */
+.global __metal_before_start
+.type __metal_before_start, @function
+__metal_before_start:
```
keith-packard, line 31: you'll need to save and restore 'ra' across this function.

bsousi5: This get called using val though. It is a single instruction to jump and save return address: jump and link (jal)

bsousi5:
Making the start = end would be a quick work-around. I did make the function a weak, type so user may include their empty function. But have not tested that yet

keith-packard:
> This looks fine to me.
> 
> As far as startup cost is concerned, maybe disabling this could be as simple as the linker script generator setting the `start` and `end` symbols equal to each other when it knows this isn't necessary?

If we can get to compiling freedom-metal for each project, instead of creating a library, then we can put conditional compilation based on project configuration which will make things like this really easy.

PR #242, opened 2020-04-06
Enable vector if platform has v-extension

rednoah91:


knightsifive:
This functionality has been added in https://github.com/sifive/freedom-metal/pull/284

PR #243, opened 2020-04-06
Add support for RISCV hardware performance counter.

NandkumarJoshi:
Note: mcountinhibit csr is not yet accessible. As per latest RiscV privileged spec v1.11, mcountinhibit controls which of the counters increment. Unused counters can be disabled to reduce power consumption. APIs to start / stop counters can be added, once this csr is supported.


src/hwperfmon.c
```diff
@@ -0,0 +1,325 @@
...
+
+/* Return codes */
+#define METAL_HWPERFMON_RET_OK 0
+#define METAL_HWPERFMON_RET_NOK 1
+
+/* Module enable flag */
+static int hwperfmon_enable = 0;
+
+/* Available hardware counters */
+static metal_hwperfmon_counter hwperfmon_count = 0;
```
bsousi5, line 75: These global variable won't work for MC designs.

src/hwperfmon.c
```diff
@@ -0,0 +1,325 @@
+/* Copyright 2020 SiFive, Inc */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include <metal/hwperfmon.h>
+#include <stdint.h>
+
+/* Macro to generate code within a switch case */
+#define METAL_HWPERFMON_HANDLE_SWITCH(m)                                       \
```
bsousi5, line 8: Some funky macro.  Not such a big fan of them, but don't have an alternative at the moment.

bsousi5:
Adding @keith-packard to review for his opinion and view

src/hwperfmon.c
nategraff-sifive: Any per-hart storage should be generated by `freedom-metal_header-generator`, in this case probably as members of the RISC-V cpu struct.

NandkumarJoshi: Added checks to make sure only single hart now writes to global variables.

src/hwperfmon.c
```diff
@@ -0,0 +1,356 @@
...
+    case METAL_HWPERFMON_COUNTER_##x:                                          \
+        __asm__ __volatile__("csrr %0, mhpmevent" #x : "=r"(val));             \
+        break;
+
+/* Macro to read HW performance monitor counter values */
+#if __riscv_xlen == 32
+#define METAL_HWPERFMON_GET_COUNT_REG(x)                                       \
+    case METAL_HWPERFMON_COUNTER_##x:                                          \
+        __asm__ __volatile__("csrr %0, mhpmcounter" #x "h" : "=r"(vh));        \
+        __asm__ __volatile__("csrr %0, mhpmcounter" #x : "=r"(vl));            \
```
keith-packard, line 41: Is this magically atomic somehow? If not, then an overflow from the low 32 bits into the high 32 bits can cause
the wrong value to be returned. The usual fix for this is to use a loop which reads high, reads low, reads high until the two high reads return the same value:
```
uint32_t h1, l, h2;
do {
   h1 = read_hi();
   l = read_lo();
   h2 = read_hi();
} while (h1 != h2);
```

src/hwperfmon.c
keith-packard: How will that help? Are the hwperf enable bits shared by all harts?

src/hwperfmon.c
```diff
@@ -0,0 +1,356 @@
...
+/* Available hardware counters */
+static metal_hwperfmon_counter hwperfmon_count = 0;
+
+/* Atomic access flag, place it into ITIM. */
+static int flag __attribute__((aligned(4))) __attribute__((section(".itim"))) =
+    0;
+
+__attribute__((naked)) static int metal_hwperfmon_cas(int *ptr, int old,
+                                                      int new) {
+    /* Perform atomic compare and swap */
```
keith-packard, line 83: Can you use the GCC atomic intrinsics instead of open-coding in assembly?

src/hwperfmon.c
```diff
@@ -0,0 +1,356 @@
...
+                         "li a0, 1 # Set return to failure. \n\t"
+                         "jr ra # Return. \n\t");
+#endif
+    /* Return error, if no support */
+    __asm__ __volatile__("li a0, 1 \n\t jr ra \n\t");
+    return 1;
+}
+
+int metal_hwperfmon_enable(void) {
+    /* Make sure only a single hart executes this code */
```
keith-packard, line 101: Do we need to enable atomics on all harts, or is it sufficient to enable them on a single hart?

src/hwperfmon.c
NandkumarJoshi: My bad, i mistakenly assumed that these counters are shared between harts. hwperfmon_count is now moved into metal cpu device instance, as suggested by Nate.

src/hwperfmon.c
NandkumarJoshi: Corrected.

src/hwperfmon.c
NandkumarJoshi: Removed related code.

src/hwperfmon.c
NandkumarJoshi: Fixed. All harts can enable their counters.

NandkumarJoshi:
> I have questions about how to reliably read counters on 32-bit systems, and whether we need to enable them on all HARTs in the system.
Reading 'high' twice seems to be reliable. Yes, we need to enable them on all harts.

NandkumarJoshi:
> I would preferred HWPM or HPM or PM, instead of HWPERFMON for all function-names, file-names, and enum/macros.
> Also metal_hwperfmon_enable() should be metal_hwperfmon_init().
> Otherwise LGTM

Fixed.

PR #246, opened 2020-04-10
Picolibc fixes

keith-packard:
This patch series switches metal-pico to use more functions from gloss so that picolibc and newlib-nano have more similar APIs for applications to use. These changes allow most of the software examples to build and run unchanged.


gloss/sys_clock_gettime.c
```diff
@@ -0,0 +1,54 @@
...
+#define MTIME_RATE_HZ MTIME_RATE_HZ_DEF
+#endif
+
+#ifndef MTIME_RATE_HZ
+#define MTIME_RATE_HZ 32768
+#endif
+
+#ifdef __METAL_DT_RISCV_CLINT0_HANDLE
+#define mtime_interrupt_controller __METAL_DT_RISCV_CLINT0_HANDLE
+#endif
```
nategraff-sifive, line 19: We should also have a case here for the CLIC.

gloss/sys_times.c
```diff
@@ -18,24 +16,19 @@ extern int _gettimeofday(struct timeval *, void *);
...
-    static struct timeval t0;
-    if (t0.tv_sec == 0 && t0.tv_usec == 0)
-        _gettimeofday(&t0, 0);
+    unsigned long long mcc;
+    unsigned long long timebase;
 
-    struct timeval t;
-    _gettimeofday(&t, 0);
+    metal_timer_get_timebase_frequency(0, &timebase);
+    metal_timer_get_cyclecount(0, &mcc);
```
nategraff-sifive, line 23: We shouldn't assume that we're on hart 0. I'd recommend calling `metal_cpu_get_current_hartid()` so that we get the timebase/cyclecount of the hart we're currently running on.

Reason being is that if we're on a multi-hart design, benchmarks are usually running on hart 1 to make sure we don't get slowed down by accidentally running on a "monitor" hart. Other harts, including hart 0, are usually in a `wfi` loop, which puts them to sleep and I'm pretty sure halts `mcycle`.

gloss/sys_clock_gettime.c
keith-packard: I changed the code to look like this:
```
#if !defined(mtime_interrupt_controller) && defined(__METAL_DT_RISCV_CLINT0_HANDLE)
#define mtime_interrupt_controller __METAL_DT_RISCV_CLINT0_HANDLE
#endif

#if !defined(mtime_interrupt_controller) && defined(__METAL_DT_RISCV_CLIC0_HANDLE)
#define mtime_interrupt_controller __METAL_DT_RISCV_CLIC0_HANDLE
#endif
```

gloss/sys_times.c
keith-packard: I added a call to metal_cpu_get_current_hartid, but then realized that the clock scaling could potentially overflow the 64-bit computation. I fixed that with a bit of algebra.

gloss/sys_times.c
```diff
@@ -3,7 +3,30 @@
...
+ *	v = v % den + (v / den) * den
+ *
+ * This lets us break the computation down as follows:
+ *
+ *	r = (v * num) / den
+ *	  = ((v % den + (v / den) * den) * num) / den
+ *	  = ((v % den) * num) / den + ((v / den) * den) * num / den
+ *	  = ((v % den) * num) / den + ((v / den) * num
+ *
+ * As long as num * den fits in 64 bits, then this compuation will fit
```
knightsifive, line 20: Good work, Keith!

To be pedantic, you might consider adding an obvious clarification,

> As long as num * den fits in 64 bits **and the (exact) result does too**

That is, this trick obviously cannot help if `v * num / den` overflows in exact arithmetic. (For example, take `v == 2`, `num == 2^64`, and `den == 1`.) 

There's also a typo: "compuation".

keith-packard: Thanks for catching the typo; I'll get that fixed.

We don't actually need a * num / den to fit in 64 bits -- what we need is for the function to "work" for time computations. That just requires:
```
moddiv(a - b, num, den) = moddiv(a, num, den) - moddiv(b, num, den)
```
As this function computes the exact result of a * num / den, we'll always satisfy this, so
even if the result gets truncated to 64 bits, as long as the application operates on the results correctly, it will get the right result.

A reasonable clarification might just note that truncating the result to 64 bits will work fine.

keith-packard:
> LGTM!

Thanks much! Please merge when you are able.

PR #248, opened 2020-04-17
When GCC is run with more pedantic flags, it needs 'inline' before type

keith-packard:
Move 'inline' before 'unsigned long long' instead of after to make GCC
happy.


keith-packard:
> When we ask for `-Wpedantic`, we sure do get `-Wpedantic`, don't we?

Yup. Feel free to merge at will (once travis is happy).

PR #257, opened 2020-04-30
Create GitHub Actions release automation

nategraff-sifive:


scripts/tag-submodules.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
```
bsousi5, line 2: is this needed? Don't we do this tag from freedom-e-sdk?

nategraff-sifive:
I'm still working on this one. I need to figure out how to run Doxygen in GitHub Actions.

scripts/tag-submodules.sh
nategraff-sifive: Accidentally copied in, it's been removed.

PR #258, opened 2020-05-03
Keep scrub code in init section

bsousi5:


src/scrub.S
```diff
@@ -5,6 +5,8 @@
  * Scrub memory with zero
  */
 
+/* Keep it in metal.init section along with _enter */
+.section .text.metal.init.enter
```
keith-packard, line 9: This is incorrect -- the _enter function is placed in this special section to ensure that it comes first in the text area, which is where the HW will start execution. If you want it in the same general area, you can use another section name starting with .text.metal.init, or .init. I'd probably call it .init._metal_memory_scrub

bsousi5: I did check and ensure _enter does come first, which it would since the linker script file already called that. But it is a good idea, I'll change to .text.metal.init

bsousi5:
Already address Keith comments. But happy to address any more if need. Will merge this for now.

PR #260, opened 2020-05-05
HPM fix for e20. Changed EOL type from windows to unix.

NandkumarJoshi:
Trap vector is set before accessing 'mcounteren' register. mvtec address alignment issue is fixed.


src/hpm.c
```diff
@@ -1,350 +1,356 @@
...
+
+int metal_hpm_disable(struct metal_cpu *gcpu) {
+    struct __metal_driver_cpu *cpu = (void *)gcpu;
+    uintptr_t temp, val = 0;
+
+    /* Check if pointer is NULL */
+    if (gcpu) {
+        /* Disable counter access */
+
+        __asm__ __volatile__("csrr %1, mtvec \n\t"
```
bsousi5, line 126: This would not work, as asm does not guarantee across different asm lines. Please check with @jim-wilson 

What you had previously would work. You simply need to check if there is trap during in the init() and return NOT_OK
Application code should check for init() success before calling the reset of the HPM apis. Your the hpm example, simply exit with return code 0, for design that don't support HPM, ie init failed.

jim-wilson: The compiler doesn't guarantee that it won't optimize code around asms.  But the asms are volatile which limits the optimizations that can be performed.  And the first asm stores a value into temp that is used by the third asm.  It looks like it should work.

However, it can only work if it is optimized, and the optimizer manages to put temp into a register.  Without optimization, the local temp might be allocated to a stack slot, and then the compiler will emit an instruction to load temp into a register right before the third asm, which will fail if you take a trap and jump into the asm.  You can make it safer by requiring optimization somehow.  Or maybe by adding checks to ensure that temp is in a register, though it isn't immediately obvious how to do that.

NandkumarJoshi: Combined multiple asm statements into one, to better avoid any unwanted side effects.

src/hpm.c
```diff
@@ -1,350 +1,345 @@
...
+    case METAL_HPM_COUNTER_##x:                                                \
+        __asm__ __volatile__("csrw mhpmcounter" #x ", zero");                  \
+        break;
+#endif
+
+/* Max available counters */
+#define METAL_HPM_COUNT_MAX 32
+
+/* Macro to check for instruction trap */
+#define MCAUSE_ILLEGAL_INST 0x02
```
bsousi5, line 73: Should use METAL_II_EXCEPTION_CODE instead. But this can be done in following commit

src/hpm.c
```diff
@@ -1,350 +1,345 @@
...
+        return METAL_HPM_RET_NOK;
+
+    /* Set trap exit, to handle illegal instruction trap. */
+    __asm__ __volatile__("la %0, 1f \n\t"
+                         "csrr %1, mtvec \n\t"
+                         "csrw mtvec, %0 \n\t"
+                         "csrr %0, mcounteren \n\t"
+                         "or %0, %0, %2 \n\t"
+                         "csrw mcounteren, %0 \n\t"
+                         ".align 4 \n\t"
```
bsousi5, line 220: this should be ".align 2"

src/hpm.c
```diff
@@ -1,350 +1,345 @@
...
+        return METAL_HPM_RET_NOK;
+
+    /* Set trap exit, to handle illegal instruction trap. */
+    __asm__ __volatile__("la %0, 1f \n\t"
+                         "csrr %1, mtvec \n\t"
+                         "csrw mtvec, %0 \n\t"
+                         "csrr %0, mcounteren \n\t"
+                         "and %0, %0, %2 \n\t"
+                         "csrw mcounteren, %0 \n\t"
+                         ".align 4 \n\t"
```
bsousi5, line 245: same here

PR #262, opened 2020-05-20
Remove vtable abstractions for drivers with one instance

nategraff-sifive:
Opening a draft pull request for early feedback on my progress here.


nategraff-sifive:
@keith-packard @bsousi5 This is now ready for review.

All vtables are removed except for those for:
 - Interrupt controllers
 - Clocks
 - UARTs

These remain (for now) because they have multiple backing implementations and multiple of these backing implementations are likely or guaranteed to appear in a single design. This might not be obvious for the UART drivers, but for example the `sifive,trace` device is a backing implementation which often appears in the same design as `sifive,uart0`. Therefore, we'll need to figure out a different method for removing those vtables which permit the drivers to coexist.

PR #263, opened 2020-05-22
Remove UART vtables

nategraff-sifive:
In order to remove the vtables from UART drivers, we need to change how standard I/O redirection works so that, for example, the `sifive,trace` standard out redirection can coexist with the `sifive,uart0` driver. As currently implemented, they both expose the internal UART driver API, which means that if we remove the vtables and implement the public API from both drivers, they conflict.

This solves the problem in the following steps:
1. We stop pretending that `sifive,trace` and `ucb,htif0` are UARTs, so they no longer need to implement the UART API, even partially.
2. Each driver capable of providing a standard I/O interface implements `metal_tty_*` functions, gated by a `METAL_STDOUT_<COMPAT_STRING>` flag so that at build-time the desired `metal_tty_*` implementation can be chosen to replace the default `nop` stubs.
3. We can then remove the UART vtables, since only `sifive,uart0` and `sifive,simuart0` remain as UART implementations and have no need to coexist.


nategraff-sifive:
Rebased on top of #262 so that I can continue development on the header rework on top of both of them.

PR #266, opened 2020-06-06
Improved memory scrub implementation.

NandkumarJoshi:
As per Keith's suggestion, only areas used by the application are zero-scrubbed rather than previous approach of  zeroing all available memory. Function metal_mem_scrub() enables user to scrub out any specific memory region or to write a new routine overriding '__metal_before_start'. 
Instead of adding scrub related code in sbrk(), heap is zeroed at boot. 
Following spec doc recommends scrubbing at the earliest out of reset.
https://github.com/sifive/arch-specs/blob/master/errors.adoc
In scrub.s, code section related to 'wait_scrub' is removed since harts are synced up later on in crt0.S
Memory pointed to by the tls pointer is left to the user to scrub.


src/scrub.S
```diff
@@ -11,9 +11,51 @@
 .option push
 .option norelax
 
-.type _metal_memory_scrub, @function
-_metal_memory_scrub:
-    /* Zero out memory */
+/* Function to zero-scrub specified memory */
```
keith-packard, line 14: I'd love to see the interface to this function documented in comments here.

src/scrub.S
```diff
@@ -26,10 +68,7 @@ _metal_memory_scrub:
 #endif
     ret
 
-/*
- * Initialize all memories available to zero
- * This must be call before seting up any stack(s)
- */
+/* Initialize memories to zero */
```
keith-packard, line 71: The comment about doing this before setting up the stack is still true, so it should be left in place.

src/scrub.S
```diff
@@ -42,81 +81,37 @@ __metal_before_start:
 
     la      t0, __metal_boot_hart
     csrr    a5, mhartid
+
     /* Disable machine interrupts to be safe */
```
keith-packard, line 85: I think this operation should be included in metal_memory_scrub so that function will operate correctly
if called outside of this environment. What other settings do we need to ensure that metal_memory_scrub does not cause a fault if run later in the program?

src/scrub.S
```diff
@@ -42,81 +81,37 @@ __metal_before_start:
...
+    beq     t1, t2, 1f
+    jal     __metal_memory_scrub
+1:
+    /* Zero out itim memory */
+    la      t1, metal_segment_itim_target_start
+    la      t2, metal_segment_itim_target_end
+    beq     t1, t2, 1f
+    jal     __metal_memory_scrub
+1:
+    /* Zero out heap memory */
```
keith-packard, line 110: I'm concerned that the heap may be large on some systems causing a long delay at this point in the program execution. That is why I suggested delaying this part of the memory scrub until sbrk is invoked. If metal_memory_scrub is careful about setting the system up to avoid exceptions, that should be fairly easy to manage.

src/scrub.S
NandkumarJoshi: Updated comments with function input parameter details.

src/scrub.S
NandkumarJoshi: Fixed.

src/scrub.S
NandkumarJoshi: Out of reset, bus error interrupts are masked out, with interrupts disabled. Once user enables BEU, registers a corresponding isr, it will catch ecc errors as they occur.
metal_memory_scrub now disables / restores global interrupts, which should avoid any spurious ecc error leading to bus error interrupt. It limits this function to be called only from m-mode. (writing to mstatus in u/s mode will raise exception).

src/scrub.S
NandkumarJoshi: I added it here, since current default size of heap in fe-sdk wasn't large enough to cause much of an impact on start up times. Large sizes will lead to concern.
Updated _sbrk() with a call to metal_memory_scrub(), to scrub off heap as needed. This ultimately, limits 'malloc' to work only on m-mode [acceptable constraint ?]. beu interrupts cannot be delegated to s/u mode.

NandkumarJoshi:
> I don't see where this is initializing the BSS section.
> I think we can use the copy of the data section to initialize that rather than zero'ing and then later copying over the top. That would require
> disabling error reporting during that copy, which seems fine. With that, program initialization will take no longer with scrubbing than without, except that we're now initializing the stack area to zero before starting.

bss is skipped here since it is already zeroed later on.
https://github.com/sifive/freedom-metal/blob/esat-267-ecc-scrub-fix/src/scrub.S#L103
At this line, machine interrupts are disabled, and not enabled till we reach bss initialization in Crt0.S 
While it seems logical to use data section copy for ecc memory initialization, but ECC section in core ip manual mentions that memories must be initialized to zeros.
"...It is recommended to use a DMA, if available, to write the entire SRAM or cache to zeros prior to enabling ECC reporting. If no DMA is present, use store instructions issued from the processor."

keith-packard:
Looks good to me, approved to merge.

PR #268, opened 2020-06-23
Build with segger libc v2004

issuehsu:


nategraff-sifive:
Closing in favor of #331, which was just merged.

PR #271, opened 2020-07-08
tty.c: added a nop hint to nop_putc() so that post-processing can see…

BrianGraysonSiV:
… the characters being output to the serial port.

With this change, a "customizable non-reserved NOP hint" instruction (slli x0, a0, 0x11) is used, with the character to be output in the specified register. This allows a post-processing tool to find occurrences of that instruction, grab the character value out of the register fields, and output the character.

A simple awk script suffices to parse the output files. Here's a one-liner version:
awk '/inst=.01151013/{printf "%c",strtonum("0x"substr($(NF-4),21,2));}' myfile

### DIFF
--- a/src/tty.c
+++ b/src/tty.c
@@ -44,7 +44,7 @@ int nop_putc(int c) __attribute__((section(".text.metal.nop.putc")));
 // look for this instruction, and use the value in a0 as the character to be
 // printed.
 int nop_putc(int c) {
-    asm volatile("slli x0,a0,0x11");
+    __asm__ volatile("slli x0,a0,0x11");
     return -1;
 }
 int metal_tty_putc(int c) __attribute__((weak, alias("nop_putc")));
--- a/src/tty.c
+++ b/src/tty.c
@@ -50,7 +50,7 @@ int nop_putc(int c) {
     // truly in a0, for easier post-processing, and so there is a single
     // 32-bit opcode to match against.
     // So explicitly ensure that the argument is placed into a0 first.
-    __asm__ volatile("mv a0, %0; slli x0,a0,0x11" :: "r"(c));
+    __asm__ volatile("mv a0, %0; slli x0,a0,0x11" ::"r"(c));
     return -1;
 }
 int metal_tty_putc(int c) __attribute__((weak, alias("nop_putc")));
--- a/src/tty.c
+++ b/src/tty.c
@@ -40,7 +40,9 @@ METAL_CONSTRUCTOR(metal_tty_init) {
  * provide a shim that eats all the characters so we can ensure that everything
  * can link to metal_tty_putc. */
 int nop_putc(int c) __attribute__((section(".text.metal.nop.putc")));
-int nop_putc(int c) { return -1; }
+// Use a customizable NOP hint instruction so that a post-processor parser can
+// look for this instruction, and use the value in a0 as the character to be printed.
+int nop_putc(int c) { asm volatile ("slli x0,a0,0x11"); return -1; }
 int metal_tty_putc(int c) __attribute__((weak, alias("nop_putc")));
 #pragma message(                                                               \
     "There is no default output device, metal_tty_putc() will throw away all input.")
--- a/src/tty.c
+++ b/src/tty.c
@@ -42,7 +42,10 @@ METAL_CONSTRUCTOR(metal_tty_init) {
 int nop_putc(int c) __attribute__((section(".text.metal.nop.putc")));
 // Use a customizable NOP hint instruction so that a post-processor parser can
 // look for this instruction, and use the value in a0 as the character to be printed.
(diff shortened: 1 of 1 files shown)


src/tty.c
```diff
@@ -40,7 +40,13 @@ METAL_CONSTRUCTOR(metal_tty_init) {
  * provide a shim that eats all the characters so we can ensure that everything
  * can link to metal_tty_putc. */
 int nop_putc(int c) __attribute__((section(".text.metal.nop.putc")));
-int nop_putc(int c) { return -1; }
+// Use a customizable NOP hint instruction so that a post-processor parser can
+// look for this instruction, and use the value in a0 as the character to be
+// printed.
+int nop_putc(int c) {
+    asm volatile ("slli x0,a0,0x11");
```
nategraff-sifive, line 47: Pretty sure we need you to use `__asm__` here in order to comply with `-std=c99`.

Cool trick!

BrianGraysonSiV: Ah, good catch! Fixed!

src/tty.c
```diff
@@ -40,7 +40,13 @@ METAL_CONSTRUCTOR(metal_tty_init) {
  * provide a shim that eats all the characters so we can ensure that everything
  * can link to metal_tty_putc. */
 int nop_putc(int c) __attribute__((section(".text.metal.nop.putc")));
-int nop_putc(int c) { return -1; }
+// Use a customizable NOP hint instruction so that a post-processor parser can
+// look for this instruction, and use the value in a0 as the character to be
+// printed.
+int nop_putc(int c) {
+    __asm__ volatile("slli x0,a0,0x11");
```
nategraff-sifive, line 47: Now that I think about it, it's probably better to do an extended assembly instruction like

```
__asm__ volatile("slli x0, %0, 0x11" :: "r" (c));
```

Technically we can expect the ABI to put `c` in register `a0`, but this tells the compiler exactly what we want to happen.

knightsifive:
Mentioning @ericlove @aswaterman who have both implemented this trick elsewhere.

src/tty.c
ericlove: lgtm.  I [used a different nop ](https://github.com/sifive/freedom-metal/blob/30fd755314b578c5dba8196c14533619fde261b1/src/tty.c#L44)in the perf-tools branch of freedom-metal, but never PR'd it because it felt inappropriate, but if this is accepted, I'll gladly switch to it and make my branch unnecessary.

aswaterman: Without @nategraff-sifive's fix, LTO could miscompile the code, since the compiler otherwise wouldn't know the dependence between `c` and the `asm` block.

Further constraining that rs1 be a0 would make the sed script easier to write, but that's a judgment call.

src/tty.c
```diff
@@ -40,7 +40,13 @@ METAL_CONSTRUCTOR(metal_tty_init) {
  * provide a shim that eats all the characters so we can ensure that everything
  * can link to metal_tty_putc. */
 int nop_putc(int c) __attribute__((section(".text.metal.nop.putc")));
-int nop_putc(int c) { return -1; }
+// Use a customizable NOP hint instruction so that a post-processor parser can
+// look for this instruction, and use the value in a0 as the character to be
```
knightsifive, line 44: For documentation purposes, it might be worth clarifying that "the value in `a0`" will have undergone some extensions (`char` -> `int` -> `int{32,64}_t`) and only the least significant 8 bits are relevant.

BrianGraysonSiV: Promotion from unsigned char to 32/64-bit int will leave the value unchanged, though, because it's an unsigned char.

Do you want me to explicitly AND it with 0xff before doing the slli, to make it completely clear?

BrianGraysonSiV: Here's code that works for the current output from Verilator +verbose logging:

awk '
/inst=.01151013/ {
  char=$(NF-4);
  char = "0x"substr(char,21, 2);
  char=strtonum(char);
  printf "%c", char;
}' $*

Or the one-liner version, for easier cut-and-pasting:

awk '/inst=.01151013/{printf "%c",strtonum("0x"substr($(NF-4),21,2));}' <filename>

src/tty.c
BrianGraysonSiV: Ah, okay, LTO is sufficient reason to tweak it. I'll do the silly move-to-self trick then, with a comment about working even if LTO would otherwise mess it up.

BrianGraysonSiV: @ericlove that `andi` nop is unfortunately in the nop hint space that is "reserved for future standard use." That's one minor reason why I went with the `slli` instead, so that it won't break when we/someone decide `andi x0, *` is some sort of hint.

src/tty.c
knightsifive: Agreed: unsigned char will get zero extended to 32 bits, so a (possible) subsequent sign extension to 64 bits is just another zero extension. Feel free to disregard my comment.

PR #272, opened 2020-07-10
Add support for 64bit event selectors. New APIs.

NandkumarJoshi:
Add support for 64bit event selectors for U8.
Add API to get supported no. of HPM counters.
Add API to assign user values to event selector register.
PR fix for changes requested by @BrianGraysonSiV

### DIFF
--- a/src/hpm.c
+++ b/src/hpm.c
@@ -20,6 +20,12 @@
         __asm__ __volatile__("csrw mhpmevent" #x ", %0" : : "r"(val));         \
         break;
 
+/* Macro to assign values into event selector register */
+#define METAL_HPM_ASSIGN_EVENT_REG(x)                                          \
+    case METAL_HPM_COUNTER_##x:                                                \
+        __asm__ __volatile__("csrw mhpmevent" #x ", %0" : : "r"(regval));      \
+        break;
+
 /* Macro to set values into event selector register */
 #define METAL_HPM_CLR_EVENT_REG(x)                                             \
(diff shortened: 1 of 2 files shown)


src/hpm.c
```diff
@@ -72,25 +78,21 @@
...
-#define METAL_HPM_RET_NOK 1
-
 int metal_hpm_init(struct metal_cpu *gcpu) {
     struct __metal_driver_cpu *cpu = (void *)gcpu;
 
     /* Check if counters are initialized or pointer is NULL */
     if ((gcpu) && (cpu->hpm_count == 0)) {
         metal_hpm_counter n;
-
+        uintptr_t bitmask;
```
BrianGraysonSiV, line 87: Can you change uintptr_t to uint64_t? Making it a 'ptr' type suggests that you are dealing with pointers, but these aren't pointers. This will also make the code do the right thing on a 32-bit system, as uint64_t is always 64 bits, but I am guessing on a 32-bit system, uintptr_t may be only 32 bits.

NandkumarJoshi: Event selection registers width varies as per platform(32/64bit). 
Counter registers are up to 64bit in width(even on 32bit platform).
used uintptr_t from standard defines, which tracks pointer-width (thus effectively register width) on a platform.
Replaced uintptr_t with metal_cpureg_t type which is defined as per cpu register width.

BrianGraysonSiV: Ah, good catch! This looks fine to me.

PR #275, opened 2020-07-17
Use PRCI/AON registers for FE310 Oscillator Drivers

nategraff-sifive:
Instead of depending upon the devicetree nodes for the oscillators to point at specific memory locations in the PRCI/AON to describe their control registers, just define their registers explicitly in the platform header.


keith-packard:
yeah, this looks a lot nicer. seems like there are some lessons here for how the wdog stuff should work too?

nategraff-sifive:
> seems like there are some lessons here for how the wdog stuff should work too?

Yeah, the watchdog driver will use the same base address.

PR #277, opened 2020-07-17
Add a nop hint to nop_putc()

nategraff-sifive:
Port of #271 to Freedom Metal refactor work.

Allows standard out to be extracted from an instruction trace when no
other stdio method is available.


keith-packard:
looks good to me. we should also figure out how to get putc hooked to semihosting as an option, but that's a separate adventure.

PR #281, opened 2020-07-17
Add back PMP API to Refactor

nategraff-sifive:


metal/pmp.h
```diff
@@ -18,8 +18,9 @@
  * can be found by reading the RISC-V Privileged Architecture Specification.
  */
 
-#include <metal/machine.h>
+#include <metal/generated/cpu.h>
 #include <stddef.h>
+#include <stdint.h>
 
 struct metal_pmp;
```
keith-packard, line 25: This looks unused at this point

src/pmp.c
```diff
@@ -2,19 +2,14 @@
 /* SPDX-License-Identifier: Apache-2.0 */
 
 #include <metal/cpu.h>
-#include <metal/machine.h>
+#include <metal/generated/pmp.h>
```
keith-packard, line 5: I think metal/pmp.h should include generated/pmp.h and this file would include that?

src/pmp.c
```diff
@@ -88,56 +75,51 @@ void metal_pmp_init(struct metal_pmp *pmp) {
...
-    }
-
     if (region > _pmp_regions()) {
         /* Region outside of supported range */
         return 2;
     }
 
     if (config.A == METAL_PMP_NA4 &&
-        pmp->_granularity[metal_cpu_get_current_hartid()] > 4) {
+        _granularity[metal_cpu_get_current_hartid()] > 4) {
```
keith-packard, line 110: This should be in the DTS; we shouldn't be doing run-time detection of this value. Sounds like a bug should be filed against the DTS generation?

PR #285, opened 2020-07-20
Fix warnings when -Wall -Wextra is on

nategraff-sifive:


src/drivers/riscv_cpu.c
```diff
@@ -49,21 +49,25 @@ int metal_cpu_set_mtimecmp(struct metal_cpu cpu, uint64_t time) { return -1; }
 int metal_cpu_enable_interrupts(struct metal_cpu cpu) __attribute__((weak));
```
keith-packard, line 49: Should these just have a void return type?

PR #286, opened 2020-07-20
Put back ITIM API to Refactor

nategraff-sifive:


metal/itim.h
```diff
@@ -0,0 +1,20 @@
...
+ * API for manipulating ITIM allocation
+ */
+
+/*! @def METAL_PLACE_IN_ITIM
+ * @brief Link a function into the ITIM
+ *
+ * Link a function into the ITIM (Instruction Tightly Integrated
+ * Memory) if the ITIM is present on the target device.
+ */
+#define METAL_PLACE_IN_ITIM __attribute__((section(".itim")))
```
keith-packard, line 18: Is there some reason this doesn't start with .text. ? That way targets without ITIM wouldn't need a .itim match in the linker script?

nategraff-sifive: That's a good improvement.

nategraff-sifive: Needs some linker script generator changes to make it work I think.

PR #287, opened 2020-07-20
Enable Vector support when present

bsousi5:


nategraff-sifive:
@ericlove beat you to it #284

bsousi5:
In favor of  #284

PR #289, opened 2020-07-20
Add enable/disable for the external interrupt

nategraff-sifive:
Also makes interrupt enable/disables in the CPU API return void. Since they can operate only on the current hart, this also removes the `struct metal_cpu` parameter.

### DIFF
--- a/src/drivers/riscv_plic0.c
+++ b/src/drivers/riscv_plic0.c
@@ -5,6 +5,7 @@
 
 #ifdef METAL_RISCV_PLIC0
 
+#include <metal/cpu.h>
 #include <metal/drivers/riscv_cpu_intc.h>
 #include <metal/drivers/riscv_plic0.h>
 #include <metal/generated/riscv_plic0.h>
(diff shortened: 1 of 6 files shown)


src/drivers/riscv_plic0.c
```diff
@@ -64,6 +67,19 @@ static __inline__ int __metal_plic0_disable(int hartid, int id) {
               (id >> METAL_PLIC_SOURCE_SHIFT) * 4) &=
         ~(1 << (id & METAL_PLIC_SOURCE_MASK));
 
+    /* Check if any PLIC interrupts remain enabled and simply exit
```
keith-packard, line 70: nice!

src/drivers/riscv_plic0.c
```diff
@@ -64,6 +67,19 @@ static __inline__ int __metal_plic0_disable(int hartid, int id) {
...
+     * if they are. */
+    for (int i = 0; i < METAL_RISCV_PLIC0_0_RISCV_NDEV % 32; i++) {
+        if (PLIC_REGW(METAL_RISCV_PLIC0_ENABLE_BASE +
```
keith-packard, line 73: I'm confused about this -- you're looping over 'i', but not using 'i' in the loop body?

PR #294, opened 2020-07-21
Fixes to enable the local-interrupt example in refactored Metal

nategraff-sifive:


templates/metal/interrupt_handlers.h.j2
```diff
@@ -21,7 +21,7 @@
 {% endif %}
 
 {% if global_interrupts is defined %}
-#define __METAL_NUM_GLOBAL_INTERRUPTS {{ global_interrupts.irqs|length }}
+#define __METAL_NUM_GLOBAL_INTERRUPTS {{ global_interrupts.irqs|length + 1 }}
```
keith-packard, line 24: I think this should probably have the correct number and fix up the allocation in the array instead

templates/src/interrupt_table.c.j2
```diff
@@ -58,6 +58,7 @@ const metal_interrupt_handler_t __metal_local_interrupt_table[__METAL_NUM_LOCAL_
 
 {% if global_interrupts is defined %}
 const metal_interrupt_handler_t __metal_global_interrupt_table[__METAL_NUM_GLOBAL_INTERRUPTS] = {
```
keith-packard, line 60: Make this +1 and stick a comment in explaining the unused slot 0

PR #297, opened 2020-07-22
Add metal/time.h and metal/timer.h

keith-packard:
These just wrap existing functions in a different API, so make
them static inline instead of separate functions.

### DIFF
--- a/metal/time.h
+++ b/metal/time.h
@@ -1,0 +1,26 @@
+/* Copyright 2019 SiFive, Inc */
+/* SPDX-License-Identifier: Apache-2.0 */
+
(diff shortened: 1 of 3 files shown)


metal/time.h
```diff
@@ -0,0 +1,24 @@
...
+#define _POSIX_TIMERS
+#include <time.h>
+
+/*!
+ * @file time.h
+ * @brief API for dealing with time
+ */
+
+static __inline__ time_t
+metal_time(void)
```
nategraff-sifive, line 17: Missing a doxygen comment here.

PR #314, opened 2020-08-05
Update Sifive L2 cache controller APIs.

NandkumarJoshi:
APIs to configure Sifive cache controller registers. If looks good, will implement these.
APIs don't make use of a 'handle', since there is only a single L2 cache device in a design.
Actual set of L2 performance monitor events are implementation defined.
https://docs.google.com/document/d/1MblvKHxNy2VIzGY0_bulemxrX8frqoq8oF9HcgJNRHM/edit?ts=5e6b0612#heading=h.wplye9b2hz2e


bsousi5:
@keith-packard @nategraff-sifive please provide you comments within the next day or so as this feature development are time sensitive.

metal/drivers/sifive_ccache0.h
```diff
@@ -1,22 +1,123 @@
...
-};
+/*! @brief Initialize cache controller with specified ways.
+ * @param ways Number of ways to be enabled.
+ * @return 0 If no error.*/
+int sifive_ccache0_init(uint32_t ways);
+
+/*! @brief Get cache configuration data.
+ * @param None.
+ * @return Cache controller configuration data.*/
+sifive_ccache0_config *sifive_ccache0_get_config(void);
```
nategraff-sifive, line 37: Instead of returning the pointer to a struct with these values (where does that even get allocated?), maybe just have four functions, one for each value? Especially since there doesn't appear to be a "set config" function, I don't see the need for the struct.

NandkumarJoshi:
Removed vtable references from 'cache.h'  APIs renamed as metal_l2cache_xxxx, since they are internally mapped to a L2 cache driver.  Clear differentiation between L1 / L2 cache metal APIs.

metal/drivers/sifive_ccache0.h
NandkumarJoshi: Function now takes a struct param from the user. This API returns read only static cache configuration data, struct is just more condensed way of returning this data to the user. Adding few more functions might unnecessarily add to already long list of driver functions.

metal/cache.h
```diff
@@ -11,51 +11,26 @@
...
- * @brief Initialize a cache
- * @param cache The handle for the cache to initialize
- * @param ways The number of ways to enable
- *
- * Initializes a cache with the requested number of ways enabled.
+ * @brief Initialize L2 cache controller
+ * @param None
+ * @return 0 If no error
  */
-__inline__ void metal_cache_init(struct metal_cache *cache, int ways) {
```
bsousi5, line 36: We should not delete this API, as part of a release update, which is what kiwi is.
Keep the API and its signature, but the vtable and structure can be remove.
It should then call the new APIs, in this case metal_l2cache_get_enabled_ways(), metal_l2cache_set_enabled_ways()
Also add a comment that this API will be deprecated in the next release.

metal/cache.h
```diff
@@ -11,51 +11,26 @@
...
- * @param cache The handle for the cache
- * @return The current number of enabled cache ways
+ * @brief Get the current number of enabled L2 cache ways
+ * @param None
+ * @return The current number of enabled L2 cache ways
  */
-__inline__ int metal_cache_get_enabled_ways(struct metal_cache *cache) {
-    return cache->vtable->get_enabled_ways(cache);
-}
+int metal_l2cache_get_enabled_ways(void);
```
bsousi5, line 26: Same here, keep the current API, add a comment it will be deprecated and call the new API

metal/cache.h
```diff
@@ -11,51 +11,26 @@
...
 /*!
- * @brief Enable the requested number of cache ways
- * @param cache The handle for the cache
- * @param ways The number of ways to enabled
+ * @brief Enable the requested number of L2 cache ways
+ * @param ways Number of ways to enable
  * @return 0 if the ways are successfully enabled
  */
-__inline__ int metal_cache_set_enabled_ways(struct metal_cache *cache,
-                                            int ways) {
```
bsousi5, line 56: Same here, keep the current API, add a comment it will be deprecated and call the new API

src/cache.c
```diff
@@ -1,13 +1,39 @@
...
+/* Copyright 2020 SiFive, Inc */
 /* SPDX-License-Identifier: Apache-2.0 */
 
 #include <metal/cache.h>
 #include <metal/machine.h>
 
-extern __inline__ void metal_cache_init(struct metal_cache *cache, int ways);
-extern __inline__ int metal_cache_get_enabled_ways(struct metal_cache *cache);
-extern __inline__ int metal_cache_set_enabled_ways(struct metal_cache *cache,
-                                                   int ways);
```
bsousi5, line 10: Keep

NandkumarJoshi:
Closing this request. Raised new one against v20.05.01-branch.
https://github.com/sifive/freedom-metal/pull/324

PR #319, opened 2020-08-12
scrub: do not include a symbol which does not exist

terpstra:
This is breaking downstream repositories

### DIFF
--- a/src/scrub.S
+++ b/src/scrub.S
@@ -37,8 +37,8 @@ __metal_before_start:
     /* Save caller ra */
     mv      s0, ra
 
-    la      t0, __metal_eccscrub_bit
-    beqz    t0, skip_scrub
+//    la      t0, __metal_eccscrub_bit
+//    beqz    t0, skip_scrub
 
     la      t0, __metal_boot_hart
     csrr    a5, mhartid


terpstra:
Or are we supposed to be providing this symbol downstream?

Help needed.

terpstra:
Maybe this should be a weak symbol?

knightsifive:
I believe this symbol is `PROVIDE`-ed here: https://github.com/sifive/ldscript-generator/blob/64d212b6c579da1a5e211fba41de2926ff880069/templates/base.lds#L89 . Personally I don't use this feature.

knightsifive:
... but yeah, should be a weak symbol IMO. Otherwise wouldn't it be undefined when using
https://github.com/sifive/ldscript-generator/blob/64d212b6c579da1a5e211fba41de2926ff880069/templates/scratchpad.lds#L15 ?
@bsousi5

bsousi5:
@terpstra the right way is to fix in the ldscript-generator. Something like this:
diff --git a/templates/base.lds b/templates/base.lds
index f2127ed..35d458b 100644
--- a/templates/base.lds
+++ b/templates/base.lds
@@ -97,6 +97,8 @@ SECTIONS
     PROVIDE( metal_{{ memory.name }}_memory_start = {{ memory.base_hex }} );
     PROVIDE( metal_{{ memory.name }}_memory_end = {{ memory.base_hex }} + {{ memory.length_hex }} );
 {% endfor %}
+{% else %}
+PROVIDE(__metal_eccscrub_bit = {{ default(0) }});
 {% endif %}

How can I reproduce this?
If you need to get pass this, you can add weak symbol in freedom-metal.

terpstra:
It looks like all I needed to do is bump ldscript-generator. Thank you for the pointer.

PR #320, opened 2020-08-13
Add Wake API for Freedom Metal code generation and installation

nategraff-sifive:
This series adds support for running Freedom Metal code generation in a wit/wake build environment.

In order to make this work, generated paths in metal.mk are now relative to the `ROOT_DIR` Make variable. This allows the delivered make fragment to be reusable outside of the Wake build environment.

### DIFF
--- a/.github/workflows/wake.yml
+++ b/.github/workflows/wake.yml
@@ -1,0 +1,27 @@
+name: wake-api-test
+
+on:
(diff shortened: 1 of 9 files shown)


nategraff-sifive:
I'll merge without review on Wednesday if I don't get any feedback on this Wake API.

PR #321, opened 2020-08-13
Add Apache license headers to new files

keith-packard:

### DIFF
--- a/metal/drivers/fixed_clock.h
+++ b/metal/drivers/fixed_clock.h
@@ -1,4 +1,20 @@
 /* Copyright 2020 SiFive, Inc. */
+/* Copyright © 2020 Keith Packard */
+/*
+ * Licensed under the Apache License, Version 2.0 (the “License”); you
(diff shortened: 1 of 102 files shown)


sifive-eblot:
Is it worth adding the full license text if the [SPDX](https://spdx.org/licenses/) line is defined?
We have the same question for scl-metal.

keith-packard:
> Is it worth adding the full license text if the [SPDX](https://spdx.org/licenses/) line is defined?

Yes, the SPDX line is not really sufficient to describe the license as
it depends on an external (non-SiFive) entity providing that
information.

I've been advised by lawyers at other companies to stick the license
header in each file to ensure the license is legally bound to the
file.

> We have the same question for scl-metal.

I'd encourage you to follow the same practice; it's a minor
inconvenience when creating a new file, but doesn't really have any
significant impact for ongoing development.

sifive-eblot:
> I'd encourage you to follow the same practice

Ok, thanks.

PR #322, opened 2020-08-14
Support Handling each PLIC interrupt on each hart on multi-core

aidenschoi:
Hi,
I have an issue.

The implementation of freedom-metal should be considered about PLIC interrupt handling on multi-core. But, currently, PLIC can be handled on only 1 core that registers an interrupt at first among several cores. For example, when the PLIC interrupt of BTN0 is registered on hart0 and PLIC of BTN1 is registered on hart1, only BTN0 works on hart0 and BTN1 doesn't work. Because mtvec and mie of only one core are enabled normally and other cores keep initial value - early_trap_vector (mtvec) and NULL (mie) instead of the address of __metal_exception_handler() (mtvec) and 0x800 (MEIE in mie).

So. I modified 2 files for this problem.
The modification in riscv_cpu.c is to enable mtvec on each core.
The modification in riscv_plic0.c is to set mie to 0x800 (MEIE) on each core.

Please refer [IPOBHD-176]( https://sifive.atlassian.net/servicedesk/customer/portal/64/IPOBHD-176) for detail.

Thanks,
Aiden.

### DIFF
--- a/src/drivers/riscv_cpu.c
+++ b/src/drivers/riscv_cpu.c
@@ -466,21 +466,21 @@ void __metal_driver_riscv_cpu_controller_interrupt_init(
             intc->metal_exception_table[i] = __metal_default_exception_handler;
         }
 
-        /*
-         * Set the real trap handler if the value of mtvec is equal to
-         * early_trap_vector. If mtvec is not equal to early_trap_vector,
-         * that means user has own trap handler, then we don't overwrite it.
-         */
-        uintptr_t mtvec;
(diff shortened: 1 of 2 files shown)


keith-packard (commented):
The issue underlying this bug is that Freedom Metal does not correctly support CPU interrupt configuration in multi-HART systems. Configuration of the CPU interrupt controller can only be done by instructions executing on the target CPU. Freedom Metal has no way to get a particular core to execute those instructions. This patch does not solve that problem. It may make some code work in some environments which does not work today, but that's essentially entirely accidental.

PR #327, opened 2020-08-25
Fix integer truncation on abs() call

nategraff-sifive:
GCC 10 complains about possible integer truncation when passing a long int to abs().


knightsifive:
Since the operands are of type `long`, better to use `labs`.

PR #329, opened 2020-08-27
Removed unnecessary use of Locks.

NandkumarJoshi:


nategraff-sifive:
How do we intend for applications to call the API? Removing all the locks isn't necessarily the right answer. Depending on what kind of concurrent access we intend, the locks around read-mutate-write steps might still be necessary.

NandkumarJoshi:
> How do we intend for applications to call the API? Removing all the locks isn't necessarily the right answer. Depending on what kind of concurrent access we intend, the locks around read-mutate-write steps might still be necessary.

_init() / _set_enabled_ways would be usually handled by a single boot hart during start-up. apart from these two APIs, rest apparently don't need any locks. also, if multiple harts try to set/clear same counter, those count values would be of no use.

PR #333, opened 2020-09-03
Fix old-style function declarations/definitions

keith-packard:
Functions taking no arguments should be defined using (void) instead
of () as the latter means 'unknown arguments' instead of 'no arguments'.

There was only one function declared using () which took a parameter,
__metal_exception_handler, which is defined to take 'riscv_xlen_t
mcause' but declared using ().


nategraff-sifive (approved):
LGTM

PR #338, opened 2020-09-09
[next] Add support for ucb,htif0 for the Spike simulator

nategraff-sifive:
In addition to adding back the HTIF driver, we also add a weak alias for metal_tty_getc() so that targets without stdio can build with picolibc.


keith-packard (approved):
Lgtm. Thanks for making this work!

PR #340, opened 2020-09-15
Remove picolibc support

nategraff-sifive:


paul-walmsley-sifive:
I'm not an expert in this code, but from a brief glance, it looks reasonable to me.  The overall direction is aligned with SiFive's plan here

paul-walmsley-sifive (approved):

PR #344, opened 2020-09-25
Fix for PLIC interrupt issue on multiple core

Veronica-Woo:
#### Issue Description
Related ticket: https://sifive.atlassian.net/servicedesk/customer/portal/64/IPOBHD-190

On U74 MC, there are 4 U74 Cores.
The external interrupt on PLIC was correctly working for hart 0.
And the PLIC interrupt for the other harts are not working at all

#### Root Cause
Wrong interrupt enable PLIC register offset was given on enabling interrupts for all hart except hart 0.
it gives the hard id on`__metal_plic0_enable` but this should be context id.

#### Bench Test
- [x] : External interrupt for all hart worked well with sesame and standard package(U74 MC) on vcu118


keith-packard (approved):
This patch looks correct by inspection.

Veronica-Woo:
This issue  is fixed on `metal-next` 
I'm closing this PR.

PR #345, opened 2020-09-25
Fix config data read error.

NandkumarJoshi:
Fix bug in sifive_ccache0_get_config().

### DIFF
--- a/src/drivers/sifive_ccache0.c
+++ b/src/drivers/sifive_ccache0.c
@@ -21,13 +21,7 @@
 #define REG_SHIFT_16 16
 #define REG_SHIFT_24 24
 
-/* Masks to get cache configuration data */
-#define SIFIVE_CCACHE0_CONFIG_BANKS_MASK 0x000000FFUL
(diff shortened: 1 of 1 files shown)


src/drivers/sifive_ccache0.c
```diff
@@ -56,21 +50,20 @@ void sifive_ccache0_get_config(sifive_ccache0_config *config) {
         val = REGW(METAL_SIFIVE_CCACHE0_CONFIG);
 
         /* Populate cache configuration data */
-        config->num_bank = (val & SIFIVE_CCACHE0_CONFIG_BANKS_MASK);
-        config->num_ways =
-            (val & SIFIVE_CCACHE0_CONFIG_WAYS_MASK) >> REG_SHIFT_8;
+        config->num_bank = (val & SIFIVE_CCACHE0_BYTE_MASK);
```
bsousi5, line 53: Adding a comment describing what being done here would help the reader, 
given that there is a lot bit manipulation (shifting and anding)

NandkumarJoshi: added a comment, describing the operation.

bsousi5 (approved):
LGTM

PR #347, opened 2020-09-29
Remove trap vector update/restore.

NandkumarJoshi:


src/drivers/sifive_l2pf0.c
```diff
@@ -32,63 +30,45 @@
...
     int hartid;
     __asm__ volatile("csrr %0, mhartid" : "=r"(hartid));
 
-    SIFIVE_L2PF0_TRAP_CAPTURE(exit, mtvec);
-
-    uint32_t val = REGW(METAL_SIFIVE_L2PF0_BASIC_CONTROL);
-
-    /* Enable L2 prefetch unit for current hart */
-    val |= REG_MASK_BITWIDTH1;
+    if (hartid < l2pf_base_len) {
```
bsousi5, line 41: You are assuming the hartID are continuous and in sequence, which it may not.
But then the generator is assuming this too. Ideal a comment should be added onthis assumption, starting with "Alert!"

bsousi5 (approved):
Otherwise LGTM

PR #349, opened 2020-10-03
Add check for a valid l2pf base address.

NandkumarJoshi:


bsousi5 (approved):
LGTM

bsousi5:
This fixes is to address Asymmetric MC design where hart0 may or may not have L2PF while other harts has L2PF.

PR #350, opened 2020-10-09
wake: use 'here' to provide base filesystem directory

mmjconolly:
Use wake's `here` self-reference variable to determine the relative path to the root of the freedom-metal repository

### DIFF
--- a/build.wake
+++ b/build.wake
@@ -50,6 +50,8 @@ global tuple MakeElfOptions =
   global IncludeDirs:  List String
   global ElfFile:      String
 
+def freedomMetalRoot = here
+
 global def runFreedomMetalInstall options =
   def outputDir = options.getFreedomMetalConfigureOptionsOutputDir
   def metalHeader = options.getFreedomMetalConfigureOptionsMachineHeader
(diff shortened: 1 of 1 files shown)


build.wake
```diff
@@ -60,20 +61,20 @@ global def runFreedomMetalInstall options =
 
   def installedFreedomMetal =
     def inputs =
-      sources "freedom-metal" `.*`
-      | filter (!matches `freedom-metal/doc/.*` _.getPathName)
+      sources freedomMetalRoot `.*`
+      | filter (!matches "{freedomMetalRoot}/doc/.*".quote _.getPathName)
```
richardxia, line 65: I think the `quote` needs to be called on a subset of the string that exclude the regular expression special characters like `.` and `*` because otherwise they would get escaped ("quoted") as literal characters. You can quote this variable part and then call [`regExpCat`](https://github.com/sifive/wake/blob/v0.19.1/share/wake/lib/core/regexp.wake#L28) in order to concatenate two RegExp objects, e.g. ``regExpCat freedomMetalRoot.quote `/doc/.*` ``.

I'm not 100% certain about this, so you might want to test out this behavior.

mmjconolly: Now using `stringToRegExp`

mmjconolly:
This now also removes `setPlanLocalOnly True` from the wake plan

build.wake
mmjconolly: Actually that returns a `Result` for a reason I can't understand

nategraff-sifive (approved):

PR #354, opened 2020-10-13
[next] Fix PLIC driver

nategraff-sifive:
Fixes:
- PLIC interrupt enable offsets now calculate the appropriate offset by hart id
- PLIC constructor does not attempt to initialize local interrupt controllers (because it can't)
- _start initializes the local interrupt controller


src/drivers/riscv_plic0.c
```diff
@@ -30,36 +30,45 @@
 extern metal_interrupt_handler_t __metal_global_interrupt_table[];
 
 static __inline__ unsigned int __metal_plic0_claim_interrupt(uint32_t hartid) {
-    return PLIC_REGW(PLIC_CONTEXT_BASE(hartid) +
-                     METAL_RISCV_PLIC0_CONTEXT_CLAIM);
+    return PLIC_REGW(
```
keith-packard, line 32: Would be easier to read if there were a macro which took the hartid and returned the context address, or maybe a macro that took a hartid and returned the context claim register address? Something to avoid duplicating this computation several times.

src/drivers/riscv_plic0.c
```diff
@@ -30,36 +30,45 @@
...
+        METAL_RISCV_PLIC0_CONTEXT_CLAIM);
 }
 
 static __inline__ int __metal_plic0_enable(int hartid, int id) {
     if (id >= METAL_RISCV_PLIC0_0_RISCV_NDEV) {
         return -1;
     }
 
+    const uintptr_t id_offset = (id / 32) * 4;
+    const uint32_t enable_mask = (1 << id % 32);
```
keith-packard, line 66: parens here to avoid confusion over precedence (yeah, I know it's correct as-is, still made me pause :-)

src/drivers/riscv_plic0.c
```diff
@@ -30,36 +30,45 @@
...
+
     PLIC_REGW(METAL_RISCV_PLIC0_ENABLE_BASE +
-              (hartid * METAL_RISCV_PLIC0_ENABLE_PER_HART) +
-              (id >> METAL_PLIC_SOURCE_SHIFT) * 4) |=
-        (1 << (id & METAL_PLIC_SOURCE_MASK));
+              (PLIC_CONTEXT_ID(hartid) * METAL_RISCV_PLIC0_ENABLE_PER_CONTEXT) +
```
keith-packard, line 69: macros here to compute this address would be helpful

gloss/crt0.S
```diff
@@ -142,6 +142,10 @@ _start:
 
 _skip_init:
 
+  /* All harts initialize their local interrupt controller */
+  csrr a0, mhartid
```
keith-packard, line 146: this passes mhartid, but riscv_cpu_intc_init receives intc. Probably best to just pass nothing and have intc_init figure things out itself.

keith-packard (changes requested):
Looking good; no major changes needed, cosmetic fixes for the plic driver if you like, then fixing the parameters passed to riscv_cpu_intc_init

PR #357, opened 2020-10-19
metal_cpu_get_current_hartid returns uint32_t

nategraff-sifive:
This fixes a warning in the asserts added to the riscv,cpu-intc driver.


keith-packard (approved):
lgtm!

PR #358, opened 2020-10-22
Inline dcache flush and discard functions

nategraff-sifive:
This eliminates the chance that (especially the discard operation)
corrupts the current stack frame and causes an infinite loop by re-using
an old return address.


keith-packard (approved):
LGTM.

PR #360, opened 2020-11-06
freedom-metal-next version of HCA

e-puerto:
freedom-metal-next driver for hca. It will manage DTS contain string like `compatible = "sifive,hca0";`


sifive-blocks/templates/metal/platform/metal_platform_sifive_hca0.h.j2
```diff
@@ -0,0 +1,30 @@
...
+    {% endif %}
+{% endfor %}
+
+#define METAL_SIFIVE_HCA0
+
+typedef enum
+{
+    HCA_AES_MODE = 0,
+    HCA_SHA_MODE = 1
+} hca0_mode_t;
```
nategraff-sifive, line 26: I don't see where this enum type is used, do we need it? If we do, maybe it should live in a static header file instead of a generated one, since I also don't think it's part of the "platform" header content.

sifive-blocks/templates/metal/private/metal_private_sifive_hca0.h.j2
```diff
@@ -0,0 +1,245 @@
...
+#define HCA_DMA_LEN_LEN_Msk               0xFFFFFFFU                               /**< HCA DMA_LEN: Length (number of 128-bits blocks) Mask */
+
+/* HCA DMA source address */
+#define HCA_DMA_SRC_SRC_Pos               0U                                       /**< HCA DMA_SRC: Source Address Position */
+#define HCA_DMA_SRC_SRC_Msk               0xFFFFFFFFU                              /**< HCA DMA_SRC: Source Address Mask */
+
+/* HCA DMA destination address */
+#define HCA_DMA_DEST_DEST_Pos             0U                                       /**< HCA DMA_DEST: Destination Address Position */
+#define HCA_DMA_DEST_DEST_Msk             0xFFFFFFFFU                              /**< HCA DMA_DEST: Destination Address Mask */
+
```
nategraff-sifive, line 217: All of the structures/types/defines above this point look like they might be better placed directly in the driver itself or in a driver-internal header file. Unless we're expecting to need to dynamically generate any of the above based on Devicetree content, I'd rather keep just the Devicetree-derived content in templates.

This same comment applies to metal_private_sifive_hca0_aes.h.j2 and the like, which don't appear to have any Devicetree-derived content in them.

templates/metal/private/metal_private_hca0.h.j2
```diff
@@ -0,0 +1,17 @@
+#ifndef __METAL_DT_HCA0__H
```
nategraff-sifive, line 1: If we're keeping this default driver support for the HCA, we should probably remove the '0' version number from the "HCA0" in the name of this file and associated content.

nategraff-sifive (changes requested):
This is great Emmanuel! I just have a few requested changes, mostly about trying to keep the templated source code to a minimum for clarity and organization.

sifive-blocks/templates/metal/platform/metal_platform_sifive_hca0.h.j2
e-puerto: It is used into sifive_hca0_aes_cipher, sifive_hca0_aes_auth_init, sifive_hca0_sha_init even if in functio sifive_hca0_aes_cipher, sifive_hca0_aes_auth_init it is commented (because set to 0 the field when you just reset it just before is not needed)

templates/metal/private/metal_private_hca0.h.j2
e-puerto: version hca1 will be coming with additional functionality.

sifive-blocks/templates/metal/private/metal_private_sifive_hca0.h.j2
e-puerto: register information from the HCA have been into sifive-block/metal/drivers directory

nategraff-sifive (approved):
Looks great to me!

PR #361, opened 2020-11-13
Fix comparator_count and compare_width fields in dt_pwm_data for PWM

aidenschoi:
**Issue Description**
PWM doesn't work well with example-pwm in freedom-metal-next branch.
There are missing comparator_count and compare_width fields in dt_pwm_data for PWM.

**Root Cause**
In core.dts, pwm node has following properties.
sifive,ncomparators
sifive,comparator-widthbits
But, j2 file includes the wrong item name. It needs to fix.

**Test**
example-pwm works well on sesame with vcu-118 fpga.

**Question**
I have another issue.
I'd like to test PWM interrupt, too. But it has only one interrupt_id in dt_pwm_data, though interrupts property in pwm node in core.dts file has 4 interrupt ids. Also, in sifive_pwm0.c, there is no API to get which interrupt index to enable/disable PWM interrupt.
I think it should also fix. Can I add the code for these issues?


nategraff-sifive (approved):
Thanks! Good catch!

PR #362, opened 2020-11-13
Port L2 features from v20.08 release to Freedom Metal next

nategraff-sifive:
Includes:
- L2 prefetcher
- L2 LIM allocation
- L2 performance monitor support


sifive-blocks/metal/drivers/sifive_l2pf0.h
```diff
@@ -0,0 +1,78 @@
...
+ * @param None.
+ * @return None.*/
+void sifive_l2pf0_disable(void);
+
+/*! @brief Get currently active L2 prefetcher configuration.
+ * @param config Pointer to user specified configuration structure.
+ * @return None.*/
+void sifive_l2pf0_get_config(sifive_l2pf0_config *config);
+
+/*! @brief Enables fine grain access to L2 prefetcher configuration.
```
keith-packard, line 73: Does this function need to be called with l2pf disabled? If so, might be nice to document that here.

sifive-blocks/src/drivers/sifive_ccache0.c
```diff
@@ -5,28 +5,55 @@
...
+extern char metal_segment_lim_target_start, metal_segment_lim_target_end;
 
 void sifive_ccache0_init(uint32_t ways) {
     sifive_ccache0_set_enabled_ways(ways);
 }
 
 METAL_CONSTRUCTOR(init_sifive_ccache0) {
-    /* Get the number of available ways per bank */
-    const uint32_t ways = __METAL_GET_FIELD(
+    size_t lim_size = 0;
```
keith-packard, line 32: This appears to assume that the only lim area used by the application is for code? I'd think we'd want to allow code, data and bss to all land in the LIM. Is that even possible? Would we have to initialize the cache controller earlier (like before any data/bss use?)

sifive-blocks/src/drivers/sifive_ccache0.c
```diff
@@ -5,28 +5,55 @@
...
 
-    /* Enable all the ways */
-    sifive_ccache0_init(ways);
+    const size_t way_size = METAL_SIFIVE_CCACHE0_0_CACHE_BLOCK_SIZE *
+                            METAL_SIFIVE_CCACHE0_0_CACHE_SETS *
+                            METAL_SIFIVE_CCACHE0_0_BANK_COUNT;
+
+    uint32_t ways_to_disable = 0;
+    if (way_size != 0) {
+        ways_to_disable = lim_size / way_size;
```
keith-packard, line 47: ways_to_disable = (lim_size + way_size - 1) / way_size

keith-packard (commented):
This is looking good; I've got a couple of questions posted inline

sifive-blocks/src/drivers/sifive_ccache0.c
nategraff-sifive: I'm going to leave additional ways to use the LIM as a future todo, since this brings it in-line with v20.08.

PR #364, opened 2020-11-20
fix template file for metal_private_gpio.h

Thomas-Kim-hub:
**Description**
In metal_private_gpio.h.j2,  template variable needs "_get_" for  get_intput/output_pin.


nategraff-sifive (approved):
Thanks for the catch! LGTM

PR #365, opened 2020-12-02
fix gpio interrupt clear bit control

Thomas-Kim-hub:
**Issue Description**

I try to clear only one bit of the interrupt pending register.
But all the bits in register is cleared.

**Root Cause**

I checked "sifive_gpio0_clear_interrupt()" in sifive_gpio.c 
To clear a pending bit, there was "|= OR" operation like 

    case METAL_GPIO_INT_RISING:
        GPIO_REGW(METAL_SIFIVE_GPIO0_RISE_IP) |= (1 << pin);
    break;

**Test**

After i  replaced  "|=" operation with "=" ,  this problem is cleared on fpga vcu118

Refer to : https://sifive.atlassian.net/servicedesk/customer/portal/64/IPOBHD-308


jaehoon-sifive:
@keith-packard, @bsousi5 Can you review it?

PR #367, opened 2020-12-09
Fix crashing due to bad macro

tamood:
The function metal_switch_get crashes in many examples, due to use of __METAL_DT_MAX_BUTTONS instead of __METAL_DT_MAX_SWITCHES.
This fix alleviates the problem.


e-puerto (approved):
LGTM

PR #369, opened 2021-01-15
Add HCA1 registers header

e-puerto:


metal/drivers/sifive_defs.h
```diff
@@ -0,0 +1,27 @@
...
+ * @file sifive_defs.h
+ * @brief Common define for Sifive drivers
+ */
+
+#pragma GCC diagnostic push
+/* Accept CMSIS syntax */
+#pragma GCC diagnostic ignored "-Wreserved-id-macro"
+
+/* following defines should be used for structure members */
+#define __IM                                                                   \
```
bsousi5, line 17: Any reason why these need to be in a separate file? If only used for HCA, it should live in hca header file

metal/drivers/sifive_hca1_regs.h
```diff
@@ -0,0 +1,835 @@
...
+        uint32_t MAJORREV:4;       /**< bit:  8..11  HCA Major Revision */
+        uint32_t _reserved0:4;     /**< bit: 12..15  (reserved) */
+        uint32_t CONFIG:5;         /**< bit: 16..20  HCA Configuration */
+        uint32_t _reserved1:11;    /**< bit: 21..31  (reserved) */
+    } b;                           /**< Structure used for bit access */
+    uint32_t w;                    /**< Structure used for word access */
+} HCA_HCA_REV_Type;
+
+/* HCA HCA Revision: Patch Revision */
+#define HCA_HCA_REV_PATCHREV_Pos        0U
```
bsousi5, line 117: You should follow the current coding styles, in that all macro/#define should be all caps.
We should also uses the full word and not abbreviation, for POSITION and MASK.

metal/drivers/sifive_defs.h
e-puerto: I Move it into sifive_hca1_regs.h

metal/drivers/sifive_hca1_regs.h
e-puerto: This file is CMSIS like, so I follow the coding style of CMSIS.

PR #371, opened 2021-02-15
PDMA Driver Changes

nvgubba:
Emmanuel/Bunnaroath can you please help me to review this pull request for the PDMA Drivers


nvgubba:
> why do we need to increase this value for the DMA driver? increase this value will increase the simulation execution time.

Initially I decreased the value to 300 for Sesame simulation to be faster. By mistake i commited the file, so i reverted it back.
For Sesame i know it is not crossing 300 threshold from DTS so i made it to 300, but if the code is getting used by some other platform it may be more than 300 Interrupt lines. Then the value is not appropriate..so i reverted back to 1024 i.e max interrupts supported by PLIC

e-puerto (approved):
LGTM

e-puerto:
DO NOT MERGE IT until DMA IP is available into Federation release branch.

PR #373, opened 2021-04-20
Fix crt0 passes unaligned argv to main()

DavidSTWChen:
argv may be unaligned depend on the context of rodata, then causes application occurs unaligned fault when paring argv

### DIFF
--- a/gloss/crt0.S
+++ b/gloss/crt0.S
@@ -293,6 +293,7 @@ secondary_main:
 /* This shim allows main() to be passed a set of arguments that can satisfy the
  * requirements of the C API. */
 .section .rodata.libgloss.start
+.align 8
 argv:
 .dc.a name
 envp:


fuchingy:
@NandkumarJoshi and @bsousi5 , could you review it?

gloss/crt0.S
```diff
@@ -293,6 +293,7 @@ secondary_main:
 /* This shim allows main() to be passed a set of arguments that can satisfy the
  * requirements of the C API. */
 .section .rodata.libgloss.start
+.align 4
```
kito-cheng, line 296: I would suggest align to 8 rather than 4, consider RV64 load a XLEN data, that will use `ld` which mean require 8 byte align.

nick-knight: Agreed. 

At a higher level, I'm never quite sure where is the appropriate place to enforce alignment. See, e.g., 
https://github.com/sifive/ldscript-generator/blob/master/templates/base.lds#L184-L196
https://github.com/sifive/ldscript-generator/blob/master/templates/base.lds#L269-L294
and this comment:
https://github.com/sifive/ldscript-generator/blob/master/templates/base.lds#L263-L266

DavidSTWChen:
Change to ".align 8"

gloss/crt0.S
sifive-eblot: If I'm not mistaken:
* `.align 4` means align on 16-byte boundary
* `.align 8` means align on 256-byte boundary

https://github.com/riscv/riscv-asm-manual/blob/master/riscv-asm.md#pseudo-ops

nick-knight: Right... on X86, `.align 4` means "align to 4-byte boundary, but on ARM and RISC-V, `.align 4` means "align to 2**4-byte boundary". Personally I prefer to use `.balign` to avoid confusion.

sifive-eblot: Very true: `.align` should be avoided, as it is highly error-prone.

DavidSTWChen:
Because '.aling 8' is not aligned at 8-byte, I would like to reopen then branch to fix it.

PR #374, opened 2021-04-26
crt0: make argv 8-byte aligned

DavidSTWChen:


fuchingy (approved):

fuchingy:
@e-puerto and @bsousi5 , could you approve this PR so it can be merged.
This PR will help improve the U84 dhrystone score.

none (5.77) --> .align 8 (5.6) --> .balign 8 (5.77)

e-puerto (approved):
LGTM

paul-walmsley-sifive:
@DavidSTWChen , we can merge this, it's no problem; just one question.  Does this patch also need to be applied to the toolchain libgloss as well and sent to the toolchain team?  i.e., https://github.com/sifive/riscv-newlib/blob/sifive-newlib-2021.04-release/libgloss/riscv/crt0.S


OIn the FESDK side, we will also have to apply this one to the new release_generator_mongoose branch; we can probably take care of that on our side.

kito-cheng:
@paul-walmsley-sifive The assumption for newlib/libgloss is: argv are prepared by the execution environment like simulator, so the execution environment has responsibility to make sure that.

In another word: I think we don't need similar change on newlib/libgloss side :)

PR #382, opened 2021-05-12
Fix clang build due to declare a symbol with both .weak and .global

kito-cheng:
- A symbol is either weak and global bind, declare a symbol with .weak
   and .global will result weak binding only in binutils.

 - But in LLVM, such kind of legacy usage is forbidden.

Ref: https://reviews.llvm.org/D90108


e-puerto (approved):
LGTM

PR #385, opened 2021-06-07
Support compact code model for freedom-metal.

Nelson1225:
Hi Guys,

I got the request recently that our freedom-e-sdk cannot build compact code model for coremark and dhrystone.  If we use the following configuration in metal.customer.lds,

MEMORY
{
    itim (airwx) : ORIGIN = 0x1801000, LENGTH = 0x1000
    ram (arw!xi) : ORIGIN = 0x1000000000, LENGTH = 0x1000000000
    rom (irx!wa) : ORIGIN = 0x200000000, LENGTH = 0x1fffffff
}

Then we will get a lot of ld truncated fails for R_RISCV_GPREL relocation.  There are two issues need to be fixed as follows,

1. The parts of .LCx and .Lx labels are actually ROdata, once we move the RWdata (gp is also there) far away from text and Rodata, then the gprel range may be not enough to use.  But if we move the ROdata and RWdata together, then gprel access should be fine.  Unfortunately gcc cannot know the section configurations until the linker have parsed the linker script, so probably always use got_gprel for those cases should be the safest method.  This is toolchain's issue rather than freedom-e-sdk, so just for reference only.

2. We need to support compact patterns in freedom-metal, just like what we did in the newlib.  This PR adds those compact cases, when RISCV_CMODEL sets to `compact`.

* When to use GPREL?  Only RWdata which are placed nearly gp.
* When to use GOT_GPREL?  ROdata, since they can be placed anywhere.  Sometimes they are near to text, but sometimes they are near to RWdata, depends on which linker script is used.
* When to use pcrel?  Symbols defined in the text section. 

Thanks
Nelson

### DIFF
--- a/gloss/crt0.S
+++ b/gloss/crt0.S
@@ -41,7 +41,15 @@ _start:
    * to just "addi gp, gp, 0". */
 .option push
 .option norelax
+#ifdef __riscv_cmodel_compact
+1:
+  auipc a0, %pcrel_hi(__global_pointer__)
+  addi a0, a0, %pcrel_lo(1b)
+  ld   gp, 0(a0)
+  add  gp, gp, a0
+#else
   la gp, __global_pointer$
+#endif
 .option pop
 
   /* Stack pointer is expected to be initialized before _start */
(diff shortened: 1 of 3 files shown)


Nelson1225:
FYI, this is the only compact spec that I can find and use,
https://github.com/ebahapo/riscv-elf-psabi-doc/blob/compact/riscv-compact.md

issuehsu:
@e-puerto I see the CI failed with message: "Unable to find image 'ubuntu:16.04' locally"
Is this a reasonable failure? Or anything we do wrong for the PR?

e-puerto:
This Pull request has been rebased and fixed due to Travis issue. See PR#394

PR #388, opened 2021-06-14
Add sifive prci support for SiFive HiFive platform

zongbox:
This driver initializes some necessary PLLs.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -57,6 +57,7 @@ nobase_include_HEADERS += \
 	metal/drivers/sifive_l2pf1.h \
 	metal/drivers/sifive_local-external-interrupts0.h \
 	metal/drivers/sifive_pl2cache0.h \
+	metal/drivers/sifive_prci0.h \
 	metal/drivers/sifive_pwm0.h \
 	metal/drivers/sifive_rtc0.h \
 	metal/drivers/sifive_spi0.h \
@@ -87,6 +88,7 @@ nobase_include_HEADERS += \
 	metal/memory.h \
 	metal/nmi.h \
 	metal/pmp.h \
+	metal/prci.h \
 	metal/privilege.h \
 	metal/pwm.h\
 	metal/rtc.h \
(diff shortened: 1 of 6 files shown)


paul-walmsley-sifive (changes requested):
Hi Zong, looks like a good start.  Same comments on this PR as on the freedom-devicetree-tools pull request: the DT compatible string, and the name of the driver, should be Unmatched-specific - so, something like "sifive,unmatched-prci0"

zongbox:
Hi Paul,
The register maps doesn't conflict in unleashed and unmatched, I'd like to distinguish the unleashed and unmatched in this driver, then do different initialization. If it is good to you, I would add the implementation above, and keep this original common name for this prci driver. I wanna do the similar thing for freedom-devicetree-tools.

paul-walmsley-sifive:
> Hi Paul,
> The register maps doesn't conflict in unleashed and unmatched, I'd like to distinguish the unleashed and unmatched in this driver, then do different initialization. If it is good to you, I would add the implementation above, and keep this original common name for this prci driver. I wanna do the similar thing for freedom-devicetree-tools.

Sure, whatever is easiest.  It might be easiest to keep them separate, due to limitations in freedom-devicetree-tools.  Whatever works for you.

My suggestion about "sifive,unmatched-prci0" was rather stupid, as @e-puerto reminded me.  It should be something like "sifive,fu740-c000-prci0" instead.    For that to work, the devicetree-overlay-generator will probably need to be modified to rewrite "sifive,ux00prci0" to "sifive,fu740-c000-prci0" if it sees that the SoC is an Almond (e.g., matching on /compatible = "SiFive,FU740-C000" or whatever)

zongbox:
Changed in v2:
- Use pre-processor directive to separate the implementation for Unleashed and Unmatched by METAL_SIFIVE_FU540_C000_PRCI0 and METAL_SIFIVE_FU740_C000_PRCI0 macro, which are generated by freedom-devicetree-tools.
- The PRCI memory map of Unleashed is subset of Unmatched, so we can use this same driver for them, the freedom-devicetree-tools will generate the different register memory map offsets for them respectively.

paul-walmsley-sifive (approved):

paul-walmsley-sifive:
Thanks Zong

PR #391, opened 2021-06-16
Add L2 Stride Prefetcher support

frankchang0125:
* Add L2 Stride Prefetcher driver.
* Initialize L2 Stride Prefetcher during boot time if SPF exists.


e-puerto (approved):
LGTM

PR #393, opened 2021-06-16
Ebl/travis workaround 1

sifive-eblot:


sifive-eblot:
Just for test, useless

PR #396, opened 2021-06-17
Add ccache0 init by default if ccache0 is present.

e-puerto:
add sifive_ccache0_init by default into metal_init function.


sifive-eblot:
LGTM

PR #399, opened 2021-06-17
Initialize V-extension CSRs

### LINKED ISSUE #375: V-extension: should initialize vstart
Although it doesn't currently appear to be explicitly mandated in any specification (ISA/ABI/etc.), I feel it's the moral responsibility of the C runtime to (re)set `vstart = 0` as part of V-extension initialization. Currently there are a few ways the Freedom-Metal user can hack it in, but there are issues with each, and it really seems like the cleanest approach is just making this part of `crt0`. It's a one-line addition; I'd plop it right after [here](https://github.com/sifive/freedom-metal/blob/6ed229fb011d02b1a999b6bfb7e68fb5bfbc3e99/gloss/crt0.S#L227).

nick-knight:
When initializing the vector unit (when present), boot code should also initialize `vstart`, `vxrm`, and `vxsat`; see [spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#state-of-vector-extension-at-reset).

Closes #375, which concerns a particular problem that may arise when this is not done.

### DIFF
--- a/gloss/crt0.S
+++ b/gloss/crt0.S
@@ -260,6 +260,9 @@ _skip_init:
   csrr a5, mstatus
   ori a5, a5, 0x200
   csrw mstatus, a5
+  csrwi vstart, 0
+  csrwi vxrm, 0
+  csrwi vxsat, 0
 1:
 
   /* This is a C runtime, so main() is defined to have some arguments.  Since
--- a/gloss/crt0.S
+++ b/gloss/crt0.S
@@ -252,7 +252,9 @@ _skip_init:
   csrwi fcsr, 0
 1:
 
-  /* Check for vector extension support and enable it if found */
+  /* Check for vector extension support and enable it if found.
+   * Omit if toolchain doesn't support the vector extension. */
+#ifdef __riscv_v
   csrr a5, misa
   li a4, 0x200000
   and a5, a5, a4
@@ -263,6 +265,7 @@ _skip_init:
   vsetivli x0, 0, e8, m1, ta, ma
   csrwi vcsr, 0
 1:
+#endif
 
   /* This is a C runtime, so main() is defined to have some arguments.  Since
    * there's nothing sane the METAL can pass we don't bother with that but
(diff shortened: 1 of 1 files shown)


gloss/crt0.S
```diff
@@ -260,6 +260,9 @@ _skip_init:
   csrr a5, mstatus
   ori a5, a5, 0x200
   csrw mstatus, a5
+  csrwi vstart, 0
```
aswaterman, line 263: We are less likely to encounter X-propagation-related failures if we actually initialize the `vtype` and `vl` CSRs.  Even though what you wrote is sufficient for executing whole-register loads and stores,we will save ourselves some pain if we just do something like `vsetivli x0, 0, e8,m1,ma,ta` here.

gloss/crt0.S
```diff
@@ -260,6 +260,9 @@ _skip_init:
...
+  csrwi vxrm, 0
+  csrwi vxsat, 0
```
aswaterman, line 265: Use `csrwi vcsr, 0` here, which is smaller and more future-proof (it will initialize other random bits of state that get added down the road).

paul-walmsley-sifive (commented):
Won't this code need to be protected via the preprocessor to ensure that it's only included for configs that support vectors, to avoid toolchain errors?

nick-knight:
@paul-walmsley-sifive : good point, the assembler needs to know about the V-extension. This is a similar problem elsewhere in this file: for example, the use of `fcsr` a few lines up. Somehow that hasn't caused problems before. I'll add preprocessor guards to this shortly.

gloss/crt0.S
nick-knight: Thanks!

gloss/crt0.S
nick-knight: Thanks!

aswaterman (approved):

gloss/crt0.S
nick-knight: By the way, I'm using spaces between the `vtype` sub-fields, as agreed here: https://github.com/riscv/riscv-v-spec/pull/693

aswaterman: ack

PR #406, opened 2021-06-25
Use balign 8 for 8 series, otherwise use align 8.

lexus0:
The performance numbers drop for e76, and s76 with coremark, and dhrystone due to [freedom-metal PR-374](https://github.com/sifive/freedom-metal/pull/374).
* For coremark: e76: 5.71 -> 5.69 ;s76: 5.72 -> 5.67
* For dhrystone: e76: 3.0 -> 2.97

I understand it is necessary for u84 (5.6 -> 5.77 in dhrystone).
Therefore, I develop a mechanism to use align or balign by recognizing RISCV_SERIES.
This mechanism need [another PR in freedom-e-sdk](https://github.com/sifive/freedom-e-sdk-private/pull/113).
Please review, and merge it as well.

### DIFF
--- a/gloss/crt0.S
+++ b/gloss/crt0.S
@@ -332,7 +332,11 @@ secondary_main:
 /* This shim allows main() to be passed a set of arguments that can satisfy the
  * requirements of the C API. */
 .section .rodata.libgloss.start
+#ifdef __is_series_8
 .balign 8
+#else
+.align 8
+#endif
 argv:
 .dc.a name
 envp:


e-puerto (changes requested):
I have some difficulties understanding why align argv into crt0.S to balign 8 or align 8 (equivalent to balign 256) have some impact on coremark and dhrystone where both do not use argv.

nick-knight:
My best guess is it forces a stricter alignment on an ELF section that ends up containing this (after linking).

lexus0:
I have no idea either. I just did some binary search to check which commit make performance drop for e76, and s76.
[This PR](https://github.com/sifive/freedom-metal/pull/374) was not merged when I was targeting the 7 series numbers..

fuhle044:
We need to look at the map file for data layout. I suspect that the argv alignment is setting the default data sectiom alignment, which then affects the section combining alignments. Or the issue is actually in the linker script and it's doing a poor job with .sdsta/.data/.bss alignments.

paul-walmsley-sifive:
Hi folks, just to follow up here: we're concerned that this patch doesn't address the root cause of the problem, and that the root cause is poorly understood.  So at this time, we're not planning on merging this.  @fuhle044 can you assign somebody in your group to root-cause this issue?  Thanks

nick-knight:
FYI @fuchingy .

fuhle044:
Already under investigation.

lexus0:
hi all,

### In short

I think I will create another PR for new flags to fit `.balign 8` instead of having strange checking system to choose align or balign.

----

### details
The flag mining is something like tweak the performance based on what you have.
We start to tweak the performance  based on `.align 8`.
I think the performance drops from `.align 8` to `.balign 8` because the data layout didn't fit perfectly.


The flags that merged into master is based on `.align 8`.
I asked @DavidSTWChen that `.balign 8` is necessary for benchmark they are using,
and make more sense than `.align 8`.

I started another round to mind the flags to fit `.balign 8`.
They can meet the target but the flags are too long.
So I'm doing some reduction work right now.
I will create another PR for new flags once I finish my work.
Then we can close this PR, and related PR in fesdk-private.

issuehsu:
Drop this PR; we will resolve the performance drop issue with freedom-e-sdk-private PR#117

PR #411, opened 2021-07-08
sifive_pl2cache0: fix missing hartid init

e-puerto:


e-puerto:
ERROR in the process.

PR #419, opened 2021-08-16
Fix wrong inline assembly code in nop_putc.

zakk0610:
We need to add `a0` into clobber list.

In clang, the result is incorrect.
```
c.mv         a1, a0
c.li         a0, -1
c.mv         a0, a1
slli         zero, a0, 0x11
c.jr         ra
```
we could rewrite it and avoid additional `mv` instruction.

### DIFF
--- a/src/tty.c
+++ b/src/tty.c
@@ -43,7 +43,8 @@ int nop_putc(int c) {
     // truly in a0, for easier post-processing, and so there is a single
     // 32-bit opcode to match against.
     // So explicitly ensure that the argument is placed into a0 first.
-    __asm__ volatile("mv a0, %0; slli x0,a0,0x11" ::"r"(c));
+    register long xc asm("a0") = c;
+    __asm__ volatile("slli x0,a0,0x11" ::"r"(xc));
     return -1;
 }
 int metal_tty_putc(int c) __attribute__((weak, alias("nop_putc")));


kito-cheng (approved):
I believe it's better and correct way to make sure the value is hold in `a0`.

nick-knight (approved):
Looks good to me. 

FWIW, this is how @ericlove implemented this hack more than 15 months ago: https://github.com/sifive/freedom-metal/commit/30fd755314b578c5dba8196c14533619fde261b1#diff-bc73b80291779f6c896ddcb3e37b2ed8567475c138048c14f5318e9771340fa3

topperc:
This issue showed up from a customer https://sifive.atlassian.net/browse/S5CSD-9412

Why was this never merged?

nick-knight:
@topperc This is a public repo; non-SiFive employees won't be able to open that Jira link. 

The components of https://github.com/sifive/freedom-e-sdk, including https://github.com/sifive/freedom-metal, are no longer actively maintained. SiFive IP customers are provided with a different environment.

kito-cheng:
@paul-walmsley-sifive could you help to merge this? customer hit some bug around this, see https://sifive.atlassian.net/browse/SCT-3021

nick-knight:
FYI: @avernhetS5 .

PR #421, opened 2021-08-27
[ESAT-400] Add Mallard-specific CPU driver to program the TwinJet prefetcher

zongbox:
Add HWPF support. This patch not only supports HWPF, but also creates an initial function
for secondary harts. HWPF CSRs are per-harts, each harts should initialize them by themselves.
In addition, we improve the METAL_CPU_GET_CSR/METAL_CPU_SET_CSR macros to support CSRs using CSR numbers, it can support using `#define` macro names of CSR numbers and accessing newly added CSRs even if toolchain does not recognize newly addes CSRs by name.


metal/csr.h
```diff
@@ -18,15 +18,22 @@
  * @param crs Register label or hex value offset to read from
  * @param value Variable name of uintprt_t type to get the value
  */
+
+#ifdef __ASSEMBLY__
+#define __ASM_STR(x) x
+#else
+#define __ASM_STR(x) #x
+#endif
+
```
e-puerto, line 27: Why do we need to have this __ASM_STR(x) define for ASSEMBLY when no commit in assembly in this PR use it ?

e-puerto (changes requested):
This commit is very confusing. in freedom metal you have 2 level. first is the "metal driver", second is generic (functionnal) driver. for example you have uart part into src/drivers directory (mainly the access of the registry of a block IP, for example __metal_driver_sifive_uart0_putc). One part into src/ (where you have upper level API, for example metal_uart_putc.
I did not see something equivalent in your proposal.

metal/csr.h
zongbox: I'm not sure if these APIs are only for freedom-metal internal use or they are used for user as well. I saw that some example programs in `software/` directory invoke `METAL_CPU_GET_CSR/METAL_CPU_SET_CSR`, if they are also used for user, we might not know how people use it, but if they are only used for freedom-metal internal, it can be dropped as you mentioned. Could you please help me to point out what is the expectation on those and whether we need to remove the __ASSEMBLY__ part. Thanks.

zongbox:
> This PR is very confusing. in freedom metal you have 2 level. first is the "metal driver", second is generic (functionnal) driver. for example you have uart part into src/drivers directory (mainly the access of the registry of a block IP, for example __metal_driver_sifive_uart0_putc). One part into src/ (where you have upper level API, for example metal_uart_putc.
> I did not see something equivalent in your proposal.

Thank you for clarifying that :) Apart from the rule as you mentioned, the thought in this patch is that the HWPF is for per-CPU as we can see that `hwpf-xxx` properties are in the CPU nodes, and it seems to me that the stuff in `src/drivers` might be for peripheral such as `soc {` node in DTS, hence I had no add something in `src/drivers`, the PMP seems to be also implemented by that way. I'm afraid that I'm misunderstanding. Please correct me and let me know if we might as well change that. Thanks.

zongbox:
Changed in v2:
 - change name to twinjet from hwpf
 - re-write the driver by separating APIs and driver.
 - rebase code base

paul-walmsley-sifive:
Just left a comment in the other PR.  Is there any way to enable this as part of a Mallard CPU "driver", such that if the DT string "sifive,mallard0" is present, then the prefetcher code is used ?

zongbox:
Changed in v3:
- Change driver to mallard CPU instead of Twinjet block
- Add a cpu_specific mechanism to support each SiFive CPU specific functionality.

PR #426, opened 2021-10-22
Update the implementation of metal_clock for segger C library

issuehsu:
This patch help to fix the seggerlib clock function issue: https://sifive.atlassian.net/browse/SCT-430


issuehsu:
move to internal PR.

PR #430, opened 2021-11-08
Enable vector unit without misa.V

### LINKED ISSUE #429: Vector unit fails to initialize for Zve* targets
The following code fails to execute when using the Zve* (embedded vector) extensions, because they don't set misa.V:

https://github.com/sifive/freedom-metal/blob/e21891cad8365b21a961444adb3d40b2aab39e93/gloss/crt0.S#L255-L268

(This code should continue to work for the V extension.) 

Unfortunately, the RISC-V standard discovery mechanism hasn’t been finalized yet. In the meanwhile, @aswaterman has suggested a workaround that can be used in this case:
1) Unconditionally initialize mstatus.VS to Dirty. This is safe on any processor.
2) Read back mstatus.VS and see if it’s set to Off. If so, there is no vector unit, and you should skip initialization.

This works for both V and Zve*.

nick-knight:
This PR attempts to resolve #429 by the techniques described therein:
- switching to the `__riscv_vector` macro;
- using the mstatus.VS trick.

My implementation does introduce a subtle change in the semantics. Previously, if I'm understanding it correctly, this code implemented the following state transition of mstatus.VS (when misa.V was set):

```
 Before |  After
-----------------
    Off | Initial
Initial | Initial
  Clean |   Dirty
  Dirty |   Dirty
```

Now, this code unconditionally sets mstatus.VS to Dirty. 

I'm not sure I understand the previous Freedom-Metal behavior (nor the similar behavior of mstatus.FS). I feel like it would be most sensible to set these bits to Initial. Perhaps the assumption is that these bits are guaranteed to be Off at this point in execution, so they will always transition to Initial? I welcome the Freedom Metal developers to clarify their intention, and take over this PR.

### DIFF
--- a/gloss/crt0.S
+++ b/gloss/crt0.S
@@ -252,16 +252,20 @@ _skip_init:
   csrwi fcsr, 0
 1:
 
-  /* Check for vector extension support and enable it if found.
-   * Omit if toolchain doesn't support the vector extension. */
-#ifdef __riscv_v
-  csrr a5, misa
-  li a4, 0x200000
-  and a5, a5, a4
-  beqz a5, 1f
+  /* Check for vector extension support and enable the vector unit if found.
+   * Omit if the toolchain doesn't support any vector extension.
+   * (NB: __riscv_vector is defined for V as well as Zve*.) */
+#ifdef __riscv_vector
+  /* Unfortunately, Zve* (embedded vector extensions) do not set misa.V,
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

paul-walmsley-sifive:
Hi @nick-knight , I just noticed that this pull request should be filed against the freedom-e-sdk-private project instead.  Could you move it there?  Can discuss further on Slack to describe what changed here.  Thanks :-)

